PR #11, opened 2022-06-07
Adds jittable Position object

EiffL:
This PR implements the `Position` object from GalSim:
https://github.com/GalSim-developers/GalSim/blob/ece3bd32c1ae6ed771f2b489c5ab1b25729e0ea4/galsim/position.py#L21
and adds a couple of related functionalities to `GSObject`.

The function `parse_pos_args` in `utilities.py` is almost just a straight copy from GalSim, but instantiates the correct jax-galsim Position class.

I don't think there is dedicated test file for positions in the GalSim test suite, but positions are used all over the place, so they get implicitly tested


jax_galsim/position.py
```diff
@@ -0,0 +1,150 @@
+import jax.numpy as jnp
+
+import galsim as _galsim
+from jax._src.numpy.util import _wraps
+from jax.tree_util import register_pytree_node_class
+
+
+@_wraps(_galsim.Position)
+class Position(object):
+    def __init__(self, *args, **kwargs):
```
b-remy, line 10: GalSim does not allow to instantiate the base class. I think it is because you either use `PositionD` or `PositionI`. Maybe it can cause issues if some part of the library is expecting either `PositionD` or `PositionI` and not `Position`

jax_galsim/position.py
```diff
@@ -0,0 +1,150 @@
...
+            except KeyError:
+                raise TypeError(
+                    "Keyword arguments x,y are required for %s" % self.__class__
+                )
+            if kwargs:
+                raise TypeError("Got unexpected keyword arguments %s" % kwargs.keys())
+
+        # Make sure the inputs are indeed jax numpy values
+        self.x = jnp.asarray(self.x)
+        self.y = jnp.asarray(self.y)
```
b-remy, line 51: Here I noticed that `jnp.asarray` truncates the input to `float32`, so this might be an issue if we work with `float64` at some point.

EiffL: Does this happen even if you set the float64 flag for JAX at the initialization?

EiffL: note that I have included the following in the conftest.py:
```
# Define the accuracy for running the tests
from jax.config import config

config.update("jax_enable_x64", True)
```

b-remy: Ah it works with this config :-) I was doing my own tests and forgot this

CHANGELOG.md
```diff
@@ -4,3 +4,10 @@
 
 
 * Changes
+  * Added simplified versions of basic objects, with no support of Chromatic objects yet:
+    * `GSObjects` and `gsparams`
+    * `Sum`
+    * `Position`
```
b-remy, line 10:
```suggestion
    * `Position`, `PositionD` and `PositionI`
```

b-remy (changes requested):
Looks good to me. If you think we can allow the use of the base `Position` class, it's ready.

jax_galsim/position.py
b-remy: Related question: since we allow to use `Position`, why do we use `@register_pytree_node_class` only for `PositionD` and `PositionI`?

EiffL: You are right, the reason I'm doing it like this is because of the mechanism to flatten and unflatten the data structure that forces me to use the init function, it cannot be just a NotImplemented trigger

b-remy (approved):

ismael-mendoza (approved):
This looks good to me too! Should we merge soon @EiffL, @b-remy?

b-remy:
Hi @ismael-mendoza , thank you for your review! In my opinion, there is still an issue from these lines:

https://github.com/GalSim-developers/JAX-GalSim/blob/f38ba4cb8b1f753fd66343a9ecb559d24d62f3e3/jax_galsim/position.py#L49-L51

This jax numpy array conversion will break vmapping because `self.x` and `self.y` are converted to `jnp.ndarray` so are not longer JAX tracers... This is what #33 is supposed to fix, by just removing these lines. The purpose of these lines were to be able to change the type of the `ndarray` in `PositionD` and `PositionI`, e.g.

https://github.com/GalSim-developers/JAX-GalSim/blob/f38ba4cb8b1f753fd66343a9ecb559d24d62f3e3/jax_galsim/position.py#L146-L150

Something we cannot do because if we create a `Position` object with float arguments, `x.astype("int")` will break.

One solution could be to remove these type conversions and add warnings or errors in `PositionD` and `PositionI` if the arguments are not the expected type.

This being said, this PR passes the tests and is working for non-vmapped inputs. What I propose is to merge it and iterate on how to solve for vmapping in #33 . What do you think about that @ismael-mendoza , @EiffL ?

ismael-mendoza:
Hi @b-remy - thanks so much for your detailed response explaining the issue. I agree that this is still a problem in this PR exactly for the reason you pointed out. 

> One solution could be to remove these type conversions and add warnings or errors in PositionD and PositionI if the arguments are not the expected type.

I like this solution, but I also agree that perhaps we should merge this first and resolve this later in #33. Perhaps we should also add specific tests for vmapping?

b-remy:
Yes 100, if you look at #33, you will see vmapping tests for all classes up to image.py

b-remy:
I agree to merge as it is all correct for floating inputs. Let's continue to clean the PRs and fix vmapping separately.

ismael-mendoza:
nice, I will go ahead then...

PR #12, opened 2022-06-07
Adds jittable wcs and Transformation

EiffL:
**This PR depends on #11 and #7, it should be rebased on `main` once these PRs are merged.**

It ports some basic WCS, to make it jittable, and also implements Transformations for simple GSObjects, which makes it possible to do flux rescaling and shifting object positions.
So with this PR you can now do:
```
gal = galsim.Gaussian(sigma=1.0) * 1.5
```

Note that in implementing the WCS object, I tried a different approach than in other places, I inherit from the GalSim object instead of copy/pasting the entire code, and just redefined  the methods that need to create jax-galsim versions of the corresponding galsim objects (Essentially Positions).

I kept the CelestialWCS out of the scope for now, because for demo1.py we just need a PixelScaleWCS.


b-remy (approved):
This PR looks good to me. Some of the` wcs.py` subclasses are still missing (e.g. `UVfunction`) but I assume it's because they are not needed for demo1.py.

jax_galsim/transform.py
```diff
@@ -0,0 +1,343 @@
...
+            obj = obj.withGSParams(self._gsparams)
+
+        self._params = {
+            "obj": obj,
+            "jac": self._jac,
+            "offset": self._offset,
+            "flux_ratio": self._flux_ratio,
+        }
+
+        if isinstance(obj, Transformation):
```
ismael-mendoza, line 65: could this block break some things when vmapping or jitting?

jax_galsim/wcs.py
```diff
@@ -0,0 +1,922 @@
...
+# We inherit from the reference BaseWCS and only redefine the methods that
+# make references to jax_galsim objects.
+class BaseWCS(_galsim.BaseWCS):
+    @_wraps(_galsim.BaseWCS.toWorld)
+    def toWorld(self, *args, **kwargs):
+        if len(args) == 1:
+            if isinstance(args[0], GSObject):
+                return self.profileToWorld(*args, **kwargs)
+            else:
+                return self.posToWorld(*args, **kwargs)
```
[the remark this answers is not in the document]
ismael-mendoza: I think it's because `profileToWorld` does not explicitly use one of the new jax objects, like `posToWorld` which uses the new object `Position`.

ismael-mendoza (commented):
I'm in the process of reviewing `wcs.py` but two quick comments for now.

jax_galsim/wcs.py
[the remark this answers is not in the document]
ismael-mendoza: Hi @b-remy - my bad, I think I was thinking of something else when I wrote this. I do agree with your confusion now. I think we should also redefine `profileToWorld` since it would create a `PositionD` jax-galsim object.

jax_galsim/wcs.py
```diff
@@ -0,0 +1,922 @@
...
+        if color is None:
+            color = self._color
+        if not isinstance(image_pos, Position):
+            raise TypeError("image_pos must be a PositionD or PositionI argument")
+        return self._posToWorld(image_pos, color=color, **kwargs)
+
+    @_wraps(_galsim.BaseWCS.toImage)
+    def toImage(self, *args, **kwargs):
+        if len(args) == 1:
+            if isinstance(args[0], GSObject):
```
ismael-mendoza, line 41: One of the options in galsim here is using a `Shear` object but we don't have one yet so its not included as one of the `if` statements here. Should we add a `#TODO` so we don't forget to add this or will the tests catch it?

jax_galsim/wcs.py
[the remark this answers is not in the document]
ismael-mendoza: why not reimplement `affine` that uses `_PositionD` (which we didn't reimplement so would use `PositoinD`)?

jax_galsim/wcs.py
```diff
@@ -0,0 +1,922 @@
...
+    # Simple.  Just call _u, _v.
+    def _posToWorld(self, image_pos, color):
+        x = image_pos.x - self.x0
+        y = image_pos.y - self.y0
+        return PositionD(self._u(x, y, color), self._v(x, y, color)) + self.world_origin
+
+    def _xyTouv(self, x, y, color):
+        x = x - self.x0  # Not -=, since don't want to modify the input arrays in place.
+        y = y - self.y0
+        u = self._u(x, y, color)
```
ismael-mendoza, line 210: where is `_u` defined??

ismael-mendoza: or is `EuclideanWCS` supposed to be an 'abstract' class?

jax_galsim/wcs.py
```diff
@@ -0,0 +1,922 @@
...
+            and self.origin == other.origin
+            and self.world_origin == other.world_origin
+        )
+
+
+class LocalWCS(UniformWCS):
+    """A LocalWCS is a `UniformWCS` in which (0,0) in image coordinates is at the same place
+    as (0,0) in world coordinates
+    """
+
```
ismael-mendoza, line 408: why was `withOrigin` not reimplemented here?

jax_galsim/wcs.py
```diff
@@ -0,0 +1,922 @@
...
+#
+#########################################################################################
+
+
+@_wraps(_galsim.PixelScale)
+@register_pytree_node_class
+class PixelScale(LocalWCS):
+    def __init__(self, scale):
+        self._params = {"scale": scale}
+        self._scale = jnp.asarray(scale)
```
ismael-mendoza, line 485: this could break jit/vmap code?

jax_galsim/wcs.py
```diff
@@ -0,0 +1,922 @@
...
+        )
+
+    def _profileToImage(self, world_profile, flux_ratio, offset):
+        return _Transform(
+            world_profile,
+            (1.0 / self._scale, 0.0, 0.0, 1.0 / self._scale),
+            flux_ratio=self._scale**2 * flux_ratio,
+            offset=offset,
+        )
+
```
ismael-mendoza, line 524: there are some Shear methods missing here, I assume it's because we don't have shear objects. But why not just include it so we don't forget later? (or at least leave #TODO)

ismael-mendoza: `_shearToWorld` and `_shearToImage`

jax_galsim/wcs.py
```diff
@@ -0,0 +1,922 @@
...
+        return abs(self._det)
+
+    def getMatrix(self):
+        """Get the Jacobian as a NumPy matrix:
+
+        numpy.array( [[ dudx, dudy ],
+                      [ dvdx, dvdy ]] )
+        """
+        return jnp.array([[self.dudx, self.dudy], [self.dvdx, self.dvdy]], dtype=float)
+
```
ismael-mendoza, line 644: The function `getDecomposition` is missing here, probably bc we don't need it yet, but we should remember it somehow

ismael-mendoza (changes requested):
Ok I finished reviewing :-) some questions about some functions that were not implemented and I can't understand why. And other questions about who we will remember later to implement the `shear` functions

jax_galsim/transform.py
```diff
@@ -0,0 +1,343 @@
...
+    def __init__(
+        self,
+        obj,
+        jac=(1.0, 0.0, 0.0, 1.0),
+        offset=PositionD(0.0, 0.0),
+        flux_ratio=1.0,
+        gsparams=None,
+        propagate_gsparams=True,
+    ):
+        self._jac = jnp.asarray(jac, dtype=float).reshape(2, 2)
```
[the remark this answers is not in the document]
b-remy: Yes, and will be fixed by #33

jax_galsim/wcs.py
ismael-mendoza: to be fixed in #33

jax_galsim/wcs.py
ismael-mendoza: It's not defined in the original GalSim either so I'm going to ignore this for now @b-remy

jax_galsim/transform.py
ismael-mendoza: nope this one is OK

ismael-mendoza (approved):

PR #15, opened 2022-06-11
Adds jittable Bounds

EiffL:
**This PR should be reviewed once #12 is merged, it adds necessary tools to have traceable bounds objects.**

This is a short PR, which should be easy to review, and that simply adds the `Bounds` object, which will be pretty handy for the next PR where we start to implement Images ;-)

One notable difference with the reference GalSim code, it is not going to be possible to easily test the validity of the bounds in jitted code (with if statements), so I got rid of a few tests like:
```python
        if (self.xmin != int(self.xmin) or self.xmax != int(self.xmax) or
            self.ymin != int(self.ymin) or self.ymax != int(self.ymax)):
            raise TypeError("BoundsI must be initialized with integer values")
```

### DIFF
--- a/jax_galsim/bounds.py
+++ b/jax_galsim/bounds.py
@@ -11,10 +11,12 @@ from jax.tree_util import register_pytree_node_class
 @_wraps(
     _galsim.Bounds,
     lax_description=(
-        "The JAX implementation will not test whether the bounds are valid, isdefined is always True."
+        "The JAX implementation will not test whether the bounds are valid."
+        "This is defined as always true."
         "It will also not test whether BoundsI is indeed initialized with integers."
     ),
 )
+@register_pytree_node_class
 class Bounds(_galsim.Bounds):
     def _parse_args(self, *args, **kwargs):
         if len(kwargs) == 0:
--- a/jax_galsim/bounds.py
+++ b/jax_galsim/bounds.py
@@ -7,6 +7,7 @@ from jax._src.numpy.util import _wraps
 from jax.tree_util import register_pytree_node_class
 
 
+# The reason for avoid these tests is that they are not easy to do for jitted code.
 @_wraps(
     _galsim.Bounds,
     lax_description=(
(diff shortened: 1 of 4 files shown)


jax_galsim/bounds.py
```diff
@@ -0,0 +1,277 @@
...
+
+# The reason for avoid these tests is that they are not easy to do for jitted code.
+@_wraps(
+    _galsim.Bounds,
+    lax_description=(
+        "The JAX implementation will not test whether the bounds are valid, isdefined is always True."
+        "It will also not test whether BoundsI is indeed initialized with integers."
+    ),
+)
+class Bounds(_galsim.Bounds):
```
ismael-mendoza, line 18: need `register_pytree_node` ?

jax_galsim/bounds.py
```diff
@@ -0,0 +1,277 @@
...
+    """
+
+    _pos_class = PositionD
+
+    def __init__(self, *args, **kwargs):
+        self._parse_args(*args, **kwargs)
+        self.xmin = jnp.asarray(self.xmin).astype("float")
+        self.xmax = jnp.asarray(self.xmax).astype("float")
+        self.ymin = jnp.asarray(self.ymin).astype("float")
+        self.ymax = jnp.asarray(self.ymax).astype("float")
```
ismael-mendoza, line 208: fixed in #33

jax_galsim/bounds.py
```diff
@@ -0,0 +1,277 @@
...
+
+    _pos_class = PositionI
+
+    def __init__(self, *args, **kwargs):
+        self._parse_args(*args, **kwargs)
+        # Now make sure they are all ints
+        self.xmin = jnp.asarray(self.xmin).astype("int")
+        self.xmax = jnp.asarray(self.xmax).astype("int")
+        self.ymin = jnp.asarray(self.ymin).astype("int")
+        self.ymax = jnp.asarray(self.ymax).astype("int")
```
ismael-mendoza, line 243: also fixed in #33

ismael-mendoza (changes requested):
I think this looks good except one small change below.

jax_galsim/bounds.py
ismael-mendoza: BR: possibly we don't need it

ismael-mendoza (approved):
All looks good to me so I'm merging now

PR #21, opened 2022-06-14
WIP: Implementing drawing functions, and testing jit patterns

EiffL:
**This PR should be rebased on main once #16 is merged, it adds a basic implementation of drawImage.**

This PR is still a WIP, but where inputs are most welcome. With the content of this PR, you can do the following thing:


EiffL:
@b-remy If you want to have a look at the problem at the moment, here it is:
https://colab.research.google.com/drive/17PgExYZB-3-a4XYkmbbFt8eKl-cj4laC?usp=sharing

```python
@jax.vmap
def drawGalaxy(flux):
  # Define a galsim galaxy as the sum of two objects
  obj1 = galsim.Gaussian(half_light_radius=1.)
  obj2 = galsim.Exponential(half_light_radius=0.5)

  # Rescale the flux of one object
  obj2 = obj2.withFlux(flux)

  # Sum the two components of my galaxy
  gal = obj1 + obj2

  return gal.drawImage(nx=128, ny=128, scale=0.02, method='no_pixel')
```

This is how far I went. It actually works fine to jit/vmap if the function `drawGalaxy` doesnt have arguments

ismael-mendoza:
I finished rebasing and merging the fixes from @b-remy and @jecampagne, still this basic example  of vmapping drawing seems to fail: 

```
@jax.vmap
def drawGalaxy(flux):
  gal = galsim.Gaussian(flux = flux, half_light_radius=1.)
  return gal.drawImage(nx=128, ny=128, scale=0.02, method='no_pixel')
```

the error I get is in the `.init` of `Image` class (which I thought would not be traced) and then it fails when it tries to check the bounds. Is the problem that the bounds are potentially changing when setting up the image?

ismael-mendoza:
UPDATE: The example above and the example from @EiffL above is now working i.e. the drawing function can now be jitted/vmapped 🎉

jax_galsim/exponential.py
```diff
@@ -46,10 +46,6 @@ def __init__(self, half_light_radius=None, scale_radius=None, flux=1.0, gsparams
         else:
             super().__init__(scale_radius=scale_radius, flux=flux, gsparams=gsparams)
 
-        self._r0 = self.scale_radius
```
EiffL, line 49: I'm confused, where are these parameters stored then?

ismael-mendoza: we made them all into properties, so they are not computed at the time of `__init__`

jax_galsim/gaussian.py
```diff
@@ -57,10 +57,6 @@ def __init__(self, half_light_radius=None, sigma=None, fwhm=None, flux=1.0, gspa
         else:
             super().__init__(sigma=sigma, flux=flux, gsparams=gsparams)
 
-        self._sigsq = self.sigma**2
```
EiffL, line 60: same question, is the explanation that sigma will be stored as a variable of the parent GSObject?

jax_galsim/gsobject.py
```diff
@@ -170,6 +169,16 @@ def kValue(self, *args, **kwargs):
         kpos = parse_pos_args(args, kwargs, "kx", "ky")
         return self._kValue(kpos)
 
+    @_wraps(_galsim.GSObject.kValue)
+    def kValue(self, *args, **kwargs):
+        kpos = parse_pos_args(args, kwargs, "kx", "ky")
+        return self._kValue(kpos)
+
+    @_wraps(_galsim.GSObject.kValue)
+    def kValue(self, *args, **kwargs):
```
EiffL, line 178: why does it appear twice?

jax_galsim/wcs.py
```diff
@@ -613,7 +616,10 @@ class JacobianWCS(LocalWCS):
     def __init__(self, dudx, dudy, dvdx, dvdy):
         self._color = None
         self._params = {"dudx": dudx, "dudy": dudy, "dvdx": dvdx, "dvdy": dvdy}
-        self._det = dudx * dvdy - dudy * dvdx
+
+    @property
+    def _det(self):
```
EiffL, line 621: why are we moving all these things into properties?

EiffL (commented):
I have left a few comments I'm curious about

jax_galsim/gsobject.py
ismael-mendoza: that's a bug 😅

ismael-mendoza (approved):

PR #34, opened 2023-02-21
Add black test to CI

ismael-mendoza:
This will ensure our code in the main branch is always black formatted (since that is the style we agreed upon)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/psf/black
-    rev: 22.3.0
+    rev: 23.1.0
     hooks:
     - id: black
       language: python
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -128,13 +128,14 @@ We highly recommend installing `pre-commit`, which will take care of running Bla
 
 ```bash
 pip install pre-commit
+pre-commit install
 ```
 
 And that's all you need to do from now on.
 
 ### Documentation style
 
-JAX-GalSim follows the NumPy/SciPy format: https://numpydoc.readthedocs.io/en/latest/format.html
+JAX-GalSim follows the NumPy/SciPy format: <https://numpydoc.readthedocs.io/en/latest/format.html>
 
 However, most JAX-GalSim function will directly inherit the documentation from the reference GalSim project. We recommend avoid copy/pasting documentation, and instead using the `_wraps` utility to automatically reuse GalSim documentation:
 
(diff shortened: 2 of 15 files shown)


.pre-commit-config.yaml
```diff
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/psf/black
-    rev: 22.3.0
+    rev: 23.1.0
```
b-remy, line 3: I wonder, is there any reason for updating black version?

CONTRIBUTING.md
```diff
@@ -128,13 +128,14 @@ We highly recommend installing `pre-commit`, which will take care of running Bla
...
 pip install pre-commit
+pre-commit install
 ```
 
 And that's all you need to do from now on.
 
 ### Documentation style
 
-JAX-GalSim follows the NumPy/SciPy format: https://numpydoc.readthedocs.io/en/latest/format.html
+JAX-GalSim follows the NumPy/SciPy format: <https://numpydoc.readthedocs.io/en/latest/format.html>
```
b-remy, line 138: Could you explain what is the purpose of doing this?

b-remy (commented):
I left two questions, but this PR looks great and very helpful. Thanks @ismael-mendoza

CONTRIBUTING.md
ismael-mendoza: This is just to make the link clickable (I have a markdown formatter that adds this by default)

.pre-commit-config.yaml
ismael-mendoza: this is a good point! we should decide whether we want to pin the black version or always use the latest. 

If we pin it, this avoids each of us having conflicting black versions. If we don't, then we need to decide for a way to ensure all developers use the same, latest black version.

ismael-mendoza: This update was the result of me running `pre-commit autoupdate`

CONTRIBUTING.md
b-remy: OK, thanks

.pre-commit-config.yaml
b-remy: Let me know if I am wrong, but we should be able to do the same tests in the CIs and locally. So we need the same black version in [`.pre-commit-config.yaml`](https://github.com/GalSim-developers/JAX-GalSim/blob/main/.pre-commit-config.yaml) (for local black formatting) and in [`python_package.yaml`](https://github.com/GalSim-developers/JAX-GalSim/blob/main/.github/workflows/python_package.yaml) for CI tests. Since you need to specify a `rev` argument in the pre-commit yml, I think we just need to specify the same version in the github workflow yaml like
```
python -m pip install flake8 pytest black==23.1.0
```

ismael-mendoza: yes, that would be perfect. I will modify and push these changes

b-remy (approved):
Looks good to me :-) Thanks @ismael-mendoza

PR #44, opened 2023-07-21
Implementation of k-space convolution 

EiffL:
This PR adds the ability to perform Fourier-based PSF convolutions. Most of this code comes from @jecampagne.


EiffL:
@ismael-mendoza could you check all the add/sub/mul things  in image.py?

ismael-mendoza:
Somethings that are missing so far: 

- Adding the pixel convolution (see TODO)
- no wrapping

ismael-mendoza:
just rebased with the `pixel` branch and uncommented the TODO, it seems like pixels convolutions are working and agree with Galsim for the example here:

```
# Define a galsim galaxy as the sum of two objects
obj1 = galsim.Gaussian(half_light_radius=1.)
obj2 = galsim.Exponential(half_light_radius=0.5)

# Rescale the flux of one object
obj2 = obj2.withFlux(0.4)

# Sum the two components of my galaxy
gal = obj1 + obj2

psf = galsim.Gaussian(half_light_radius=2.)

obs_gal = Convolution(gal, psf)
```

EiffL:
Awesome :-D

jax_galsim/image.py
```diff
@@ -1016,7 +1077,7 @@ def Image_iadd(self, other):
         array = self.array + a
     else:
         array = (self.array + a).astype(self.array.dtype)
-    return Image.init(array, bounds=self.bounds, wcs=self.wcs)
+    return Image(array, _bounds=self.bounds, wcs=self.wcs, _dtype=self.dtype)
```
EiffL, line 1080:
```suggestion
    return Image(_array=array, _bounds=self.bounds, wcs=self.wcs, _dtype=self.dtype)
```

jax_galsim/image.py
```diff
@@ -1053,7 +1114,7 @@ def Image_mul(self, other):
         a = other.array
     except AttributeError:
         a = other
-    return Image.init(self.array * a, bounds=self.bounds, wcs=self.wcs)
+    return Image(self.array * a, _bounds=self.bounds, wcs=self.wcs, _dtype=self.dtype)
```
EiffL, line 1117:
```suggestion
    return Image(_array=self.array * a, _bounds=self.bounds, wcs=self.wcs, _dtype=self.dtype)
```

jax_galsim/image.py
```diff
@@ -1068,7 +1129,7 @@ def Image_imul(self, other):
         array = self.array * a
     else:
         array = (self.array * a).astype(self.array.dtype)
-    return Image.init(array, bounds=self.bounds, wcs=self.wcs)
+    return Image(array, _bounds=self.bounds, wcs=self.wcs, _dtype=self.dtype)
```
EiffL, line 1132:
```suggestion
    return Image(_array=array, _bounds=self.bounds, wcs=self.wcs, _dtype=self.dtype)
```

EiffL (commented):
Some small suggested edits

jax_galsim/convolve.py
```diff
@@ -0,0 +1,331 @@
...
+        s = "galsim.Convolve(%s" % (", ".join(str_list))
+        if self.real_space:
+            s += ", real_space=True"
+        s += ")"
+        return s
+
+    # JEC not sure if it is used
+    # def _prepareDraw(self):
+    #    for obj in self.obj_list:
+    #        obj._prepareDraw()
```
ismael-mendoza, line 212: remove this commented out code?

tests/jax/galsim/test_image_jax.py
```diff
@@ -4360,6 +4360,8 @@ def test_int_image_arith():
         full %= imd
 
 
+# TODO: add back once wrapping is implemented.
```
ismael-mendoza, line 4363: need to remove this skip

EiffL:
Well, so, for some reason there appears to be an offset in the fft drawing method, also I still am not sure whether the rotations work correctly, this should be tested

EiffL:
So, I did 2 additional things:
- I changed the way that drawImage works so that (even if it's not recommended to do so) the old expected galsim behaviour of being able to update an image will work:
https://github.com/GalSim-developers/JAX-GalSim/blob/477c2cef15a5db810484df1855dfa9dd0a40a810/jax_galsim/gsobject.py#L685

- I updated the GalSim branch we use for testing to this branch: https://github.com/GalSim-developers/GalSim/tree/fix_tests
where we can make edits like making the np.testing function array-friendly when desired, with high likelihood that the GalSim Devs will be ok with merging these small modifs. We should use that mechanism to fix tests so that they work for us, when these fixes are really minimal

EiffL:
One quick note before leaving for the airport, the reason the Fourier transform is behaving in a funny way is that I found a bug yesterday here:
https://github.com/GalSim-developers/JAX-GalSim/blob/477c2cef15a5db810484df1855dfa9dd0a40a810/jax_galsim/gsobject.py#L812

EiffL:
ok, so with the latest fixes, the drawImage (method='fft') wotks again, but I don't think it works as well as what we had the other day (which was actually buggy as it was using arrays of the wrong size, but thanks to JAX not checking for bounds, ended up working nevertheless).

We currently have only one test that fails, but it's a pretty important one which tests whether shear, rotations, and shifts work properly. For sure it should get sorted before we merge.

That's probably as far as I can go on this project for now unfortunately :-(

jax_galsim/convolve.py
EiffL: sure... not sure what prepareDraw was supposed to do

ismael-mendoza:
@EiffL  - thanks for all your work on this. Could you elaborate a bit on why you think the `fft` method is not working as well as before?

EiffL:
If you plot residuals vs galsim there are some ugly aliasing patterns

EiffL:
Ok :-) I think I fixed all the main problems. All the tests pass now, and I also added a few ones to do some basic checks against the reference GalSim in `temporary_tests.py`
It's pretty cool, you can test jax_galsim against galsim with this sort of scheme:
```python
def func(galsim):
      gal =  galsim.Gaussian(flux=1, sigma=1).shear(e1=0.1, e2=0.)
      pixel = galsim.Pixel(scale=0.2, flux=1.0)
      conv = galsim.Convolve([gal, pixel])
      return conv.drawImage(
          nx=32, ny=32, scale=0.2, method="sb").array

np.testing.assert_array_almost_equal(
      func(ref_galsim),  func(jax_galsim),
      5,
      err_msg="Using GSObject Convolve([psf,pixel]) disagrees with expected result")
```


So, anyway, I think this PR is fully ready for review @ismael-mendoza

jax_galsim/convolve.py
EiffL: I removed it

ismael-mendoza:
Amazing, thanks @EiffL ! Will review soon, very interested in hearing about the fix at some point later.

jax_galsim/image.py
```diff
@@ -561,6 +582,56 @@ def __setitem__(self, *args):
...
+            raise _galsim.GalSimValueError(
+                "Invalid value for hermitian", hermitian, (False, "x", "y")
+            )
+
+    def _wrap(self, bounds, hermx, hermy):
+        """A version of `wrap` without the sanity checks.
+
+        Equivalent to ``image.wrap(bounds, hermitian=='x', hermitian=='y')``.
+        """
+        ret = self.subImage(bounds)
```
ismael-mendoza, line 630:
```suggestion
        # NOTE: Wrapping not yet implemented.
        ret = self.subImage(bounds)
```

ismael-mendoza (commented):
It looks very good! All the tests are similarly passing for me. Although I do have some questions below but not sure if it stop merging.

@jecampagne feel free to chime in if you are able to since a lot of the code was borrowed from you. 

One last outstanding thing if we should add the draw methods for the Exponential here or in a different PR

tests/GalSim
ismael-mendoza, line 1: @EiffL did you update something in Galsim?

jax_galsim/core/draw.py
```diff
@@ -35,3 +33,44 @@ def draw_by_xValue(
...
+    im = jax.vmap(lambda *args: gsobject._kValue(PositionD(*args)))(
+        coords[..., 0], coords[..., 1]
+    )
+    im = (im).astype(image.dtype)
+
+    # Return an image
+    return Image(array=im, bounds=image.bounds, wcs=image.wcs, check_bounds=False)
+
+
+def apply_kImage_phases(gsobject, image, jacobian=jnp.eye(2)):
```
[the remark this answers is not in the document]
ismael-mendoza: Thanks @jecampagne I see now that if there is no offset, the function will do nothing as in Galsim. Thanks!

tests/GalSim
EiffL: Hi @ismael-mendoza , yes I updated to a specific galsim branch where we can push updates for the tests:
https://github.com/GalSim-developers/GalSim/tree/fix_tests

EiffL:
Thanks @jecampagne :-) I've mostly copy/pasted your code over and did a bunch of reformatting, I credited you in the corresponding commits so that your code contributions will be tracked in the project history :-)

ismael-mendoza (approved):
LGTM! I think it's ready to merge :D

PR #47, opened 2023-07-22
Restoring canonical GalSim Image API

EiffL:
This PR aims to revert some of the recent changes that have modified the API for galsim.Image. The hope here is that we can retain the original API, while making sure we can still jit and vmap when and where we need it.

### DIFF
--- a/jax_galsim/position.py
+++ b/jax_galsim/position.py
@@ -120,7 +120,7 @@ class Position(object):
             a `galsim.PositionD` instance.
         """
         shear_mat = shear.getMatrix()
-        shear_pos = jnp.dot(shear_mat, self.array)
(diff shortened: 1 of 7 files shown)


jax_galsim/position.py
```diff
@@ -116,7 +120,7 @@ def shear(self, shear):
             a `galsim.PositionD` instance.
         """
         shear_mat = shear.getMatrix()
-        shear_pos = jnp.dot(shear_mat, self.array)
+        shear_pos = jnp.dot(shear_mat, jnp.stack([self.x, self.y], axis=0))
```
EiffL, line 123:
```suggestion
        shear_pos = jnp.dot(shear_mat, self.array)
```

EiffL:
Ok, all tests on `main` now work with the original API :-)

ismael-mendoza (approved):
Looks great!

PR #57, opened 2023-09-13
PROD clean out pre-commit config and add ruff settings

beckermr:
This PR cleans up the pre-commit settings and adds ruff settings. 

It should follow #55

PR #58, opened 2023-09-13
BUG make sure pytest can run -vvs

beckermr:
This PR fixes an internal error in the pytest integration when pytest is run with `-vvs`. 

This PR should follow #57


ismael-mendoza (approved):
Thanks for fixing that!

PR #61, opened 2023-09-18
ENH add fits readers

beckermr:
This PR adds the fits readers. A bunch of associated changes had to go in as well to make this feasible.


jax_galsim/gsparams.py
```diff
@@ -21,13 +21,43 @@ class GSParams:
...
 
         In the former case, it returns gsparams, in the latter it returns default
         (GSParams.default if no other default specified).
         """
+        # we allow conversion from the original GalSim GSParams object
+        if isinstance(gsparams, _galsim.GSParams):
+            gsparams = GSParams.from_galsim(gsparams)
+        if default is not None and isinstance(default, _galsim.GSParams):
+            default = GSParams.from_galsim(default)
+
```
beckermr, line 60: Sometimes these things get injected during testing. If we'd rather have `check` error on mixing, then I'll need to go back to something else to fix this.

tests/GalSim
beckermr, line 1: I had to do a bunch of updates to the galsim test suite. I pushed to the same branch that @EiffL was using.

tests/jax/galsim/test_image_jax.py
beckermr, line 1: The changes here will need to be put into galsim.

jax_galsim/bounds.py
```diff
@@ -126,10 +128,12 @@ def includes(self, *args):
         else:
             raise TypeError("include takes at most 2 arguments (%d given)" % len(args))
 
-    def expand(self, factor):
+    def expand(self, factor, yfactor=None):
```
ismael-mendoza, line 131: any reason to have a different naming convention than the `expand` in the original Galsim? (uses `factor_x` and `factor_y`)

ismael-mendoza: The function works the same though so not very important

ismael-mendoza (approved):
LGTM! Just one minor comment

jax_galsim/bounds.py
beckermr: fixed!

PR #62, opened 2023-09-19
BUG fix accuracy loss in sheared image test

beckermr:
When I enabled the upstream galsim tests for shears, I noticed an orders of magnitude loss in accuracy for rendering images in the test. This PR fixes that.


jax_galsim/gsobject.py
```diff
@@ -719,6 +719,7 @@ def drawImage(
         image_in._array = image._array
         image_in.added_flux = image.added_flux
         image_in._bounds = image.bounds
+        image_in.wcs = image.wcs
```
beckermr, line 722: This is the actual bug. cc @rmjarvis

jax_galsim/image.py
beckermr, line 1: This file had bugs in the `__imul__` etc. methods where they did not return `self`.

jax_galsim/image.py
```diff
@@ -1345,6 +1386,6 @@ def Image_or(self, other):
 Image.__rand__ = Image_and
 Image.__rxor__ = Image_xor
 Image.__ror__ = Image_or
-Image.__iand__ = Image_and
-Image.__ixor__ = Image_xor
-Image.__ior__ = Image_or
+Image.__iand__ = Image_iand
```
beckermr, line 1389: It also didn't have the proper functions for `__iand__` etc.

beckermr:
I added the real space image wrapping but have punted on the k-space stuff. I will submit that in another PR, but it will take longer.

jax_galsim/image.py
```diff
@@ -1345,6 +1386,6 @@ def Image_or(self, other):
...
+Image.__ixor__ = Image_ixor
+Image.__ior__ = Image_ior
```
rmjarvis, line 1391: Why do this indirection?  Why not just define them in the Image class directly?

cf. https://github.com/GalSim-developers/GalSim/blob/releases/2.4/galsim/image.py#L1612

tests/jax/galsim/test_image_jax.py
```diff
@@ -4500,57 +4502,59 @@ def test_wrap():
...
-    )
-    np.testing.assert_array_equal(
-        im3_wrap.array,
-        im3[b3].array,
-        "image.wrap(%s) did not return the right subimage",
-    )
-    np.testing.assert_equal(
-        im3_wrap.bounds, b3, "image.wrap(%s) does not have the correct bounds"
-    )
+    # FIXME: turn on when hermitian wrapping is implemented
```
ismael-mendoza, line 4505: Might you want to document the remaining issue with hermitian images in an issue in addition to this FIXME?

ismael-mendoza: also if you might mention when this comes up that would be great, I'm not too familiar with when we might need to do wrapping of hermitian images

jax_galsim/image.py
ismael-mendoza: Hmm, I don't remember why we did it this way instead of the Galsim way. @EiffL @jecampagne do you remember?

jax_galsim/core/wrap_image.py
```diff
@@ -0,0 +1,35 @@
+# from functools import partial
+import jax
+import jax.numpy as jnp
+
+
+@jax.jit
+def wrap_nonhermition(im, xmin, ymin, nxwrap, nywrap):
```
ismael-mendoza, line 7: Not too familiar with wrapping myself, so @jecampagne might have suggestions/opinions here

jax_galsim/image.py
```diff
@@ -1152,10 +1162,10 @@ def Image_isub(self, other):
         a = other
         dt = type(a)
     if dt == self.array.dtype:
-        array = self.array - a
+        self._array = self.array - a
     else:
-        array = (self.array - a).astype(self.array.dtype)
-    return Image(array, bounds=self.bounds, wcs=self.wcs)
+        self._array = (self.array - a).astype(self.array.dtype)
+    return self
```
ismael-mendoza, line 1168: I think you only need to return `self` in methods that start with `i` ? like `iadd`

ismael-mendoza: e.g. Galsim `__sub__` returns a new image: https://github.com/GalSim-developers/GalSim/blob/a1f0a11a24690a279abd7027b116ec3e4986643d/galsim/image.py#L1644

ismael-mendoza (changes requested):
Thanks @beckermr , so was the accuracy loss coming from the lack of wrapping? 

I just have one requested change regarding the `__iadd__`, etc. functions  and a couple of questions.

jax_galsim/image.py
beckermr: This one is for the isub method. The github diff is showing the wrong class.

beckermr:
The accuracy loss was not wrapping but due to the WCS bug.

tests/jax/galsim/test_image_jax.py
beckermr: I added a note to the contributing document.

jax_galsim/core/wrap_image.py
beckermr: Happy to have suggestions on this. I spent a lot of time working on this specific function and this is the only version I could come up with that is differentiable and doesn't require excess recompilation.

beckermr:
Also as a quick comment. The test suite didn't catch the imul etc. bugs, either in galsim or jax-galsim. It may be that we never got the tests to run all the way to the right spot, but definitely something we should think about for future PRs.

jax_galsim/image.py
ismael-mendoza: ah gotcha, thanks!

ismael-mendoza (approved):

jax_galsim/core/wrap_image.py
jecampagne: Well, if I've fight myself to try to implement a wrapping for the case used by drawFFT, I had no success to get working a   reverse-mode differentiable version due to loop/while. Now, this wrapping is the case of hermx=hermy=False in the Galsim Image.cpp:wrapImage language. if @beckermr 1) gets it right wrt to Galsim, 2) manages to get jacrev ok then it is perfect. I guess it is working because the for_loop bouds are fixed (ie. 0 or image shape). I've nothing to say. In the code @beckermr  it is mentioned that may be you have tried different way to test 
```python
jnp.abs(ii - i) + jnp.abs(jj - j) != 0, 
```
do you see some drawbacks?
By the way where the wrapping non-hermitian is used?

beckermr: I tried a bunch of more logic like things like ii != i with jnp.all etc and got errors. 

Idk where the code uses real space wrapping. However the api supports it so I guess we should implement it.

beckermr: I should retest the gradients on this one btw. It worked for a previous version but I made a few changes and forgot to retest.

rmjarvis: > By the way where the wrapping non-hermitian is used?

hermitian='x' is the only thing that is used in practice.  I.e. other than in the unit tests of the wrap function.

The non-hermitian case was easier to code up, so I started with that.  Then once I figured out how to do the 'x' version correctly, it seemed worth doing 'y' too, so I wouldn't have to revisit it later if a case for a hermitian image stored that way ever became relevant.  But all the code that calls this have hermitian='x'.

PR #86, opened 2024-01-07
Spergel

jecampagne:
This is an implementation of the Spergel profile. 
In the Galsim code the Spergel index (nu) is restricted to [-0.85, 4.0]. In the JAX code I do not know to assert a use of non valid nu input parameter.

Notice that nu=0.5 is exactly the Exponential profile. Spergel profile might be a good alternative to the Sersic profile [as realized here](https://arxiv.org/pdf/1007.3248.pdf).

The present code pass test_spergel.py Galsim tests excepting the photon shooting even if I have setup a first version using a inversion method using JAX-Galsim bisection root algo.

A side remark. For Moffat & Spergel we need K_nu Bessel function. For the time beeing this Bessel func is defined both in moffat.py and spergel.py. I guess this is not optimal and this function would have to be defined in the core directory.


jecampagne:
I do not understand such message in the pytest used in the PR causing test failure 
```
" test_api_gsobject[docs-methods]: 
self = <[AttributeError("'Spergel' object has no attribute 'beta'") raised in repr()] Spergel object at 0x7f77ae66dd90>
```

tests/jax/test_api.py
```diff
@@ -36,6 +36,7 @@ def test_api_same():
     "Either scale_radius or half_light_radius must be specified",
     "One of sigma, fwhm, and half_light_radius must be specified",
     "One of scale_radius, half_light_radius, or fwhm must be specified",
+    "One of scale_radius, half_light_radius must be specified"
```
beckermr, line 39:
```suggestion
    "One of scale_radius, half_light_radius must be specified",
```

tests/jax/test_api.py
```diff
@@ -397,7 +398,7 @@ def test_api_gsobject(kind):
             if inspect.isclass(_attr) and issubclass(_attr, jax_galsim.GSObject):
                 classes.append(_attr)
     cls_tested = set()
-    for cls in classes:
+    for cls in classes:
```
beckermr, line 401:
```suggestion
    for cls in classes:
```

beckermr (commented):
I found a few small typos that may help get the tests to pass.

jax_galsim/spergel.py
```diff
@@ -0,0 +1,424 @@
...
+
+@jax.jit
+def fsmallz_nu(z, nu):
+    def fnu(z, nu):
+        """z^nu K_nu[z] z -> 0 O(z^4) z > 0"""
+        z2 = z * z
+        z4 = z2 * z2
+        c1 = jnp.power(2.0, -6.0 - nu)
+        c2 = _gamma(-2.0 - nu)
+        c3 = _gamma(-2.0 + nu)
```
beckermr, line 107: This variable is not used. Is that a bug?

jecampagne: Thanks, I've missed c3 coefficient.

jecampagne:
Now test_api.py have detected that my xValue is not/ jit-grad friendly.  I must investigate.

jecampagne:
Well I'm a bit puzzled but I guess this is related to 
https://jax.readthedocs.io/en/latest/faq.html#gradients-contain-nan-where-using-where

in fact fsmallz_nu cannot be used with integer nu so then the gardient. So I need to protect it according to the above doc.
The pb is that jnp.power(z, nu) * _Knu(nu, z) is not accurate for small value of z which is a pb for very sharp Spergel profile.

jecampagne:
I'm trying to investigating the pytest Failure on test_serpel.py::test_spergel do_shoot for nu>-0.3. But if I use locally at the root of JAX-Galsim:
```python
python tests/GalSim/tests/test_spergel.py
```
there is no crash/failure detected, while if I use
```python
 pytest -k test_spergel.py
```
then there is a Failure detection (the same as in the PR automatic processing). Is there a simple explanation of the differences between the two cmds?

beckermr:
The direct call doesn't replace galsim with jax_galsim.

jecampagne:
Making progress: the do_shoot test for nu>-0.3 is ok. Now there is a assert Fail later. So work is still 
needed.

jecampagne:
I have made progress to get test_serpel.py::test_spergel do_shoot working. 
Now in the test_serpel.py::test_spergel_shoot : nu >0 tests are ok. The test with
```python 
 obj = galsim.Spergel(nu=-0.6, half_light_radius=3.5, flux=1.e4)
```
fails as `added_flux` from 
```python
added_flux, photons = obj.drawPhot(im, poisson_flux=False, rng=rng.duplicate())
```
differs from `obj.flux`:
```python
obj.flux =  10000.0
added_flux =  9999.0
```
Now, looking at `SBSpergel.cpp::SpergelInfo::shoot` C++ function Galsim comment is in the context of `nu<=0`
```
     // exact s.b. profile diverges at origin, so replace the inner most circle
     // (defined such that enclosed flux is shoot_acccuracy) with a linear function
    // that contains the same flux and has the right value at r = rmin.
    // So need to solve the following for a and b:
    // int(2 pi r (a + b r) dr, 0..rmin) = shoot_accuracy
    // a + b rmin = K_nu(rmin) * rmin^nu
```
My comment is the following: 
1) The Spergel profile noted `I_{nu}(r) = xnorm(nu) * f(r/r0,nu)`  is such that
```math
$$\Large f(z,\nu)=z^\nu K_\nu(z)$$ 
```
as the following property for `z->0` (or `r->0` for `I_{nu}(r)`)
```math
$$
\begin{cases}\Large
 f(z,\nu>0) = 2^{\nu -1} \Gamma(\nu)\\
\Large f(z,\nu=0) \propto \log(1/z) \\
\Large f(z,\nu<0) \propto 1/z^{-2\nu}
\end{cases}
$$
``` 
So, I agree that `I_{nu}(r)` is diverging for `nu<=0`. 
2) But the cumulated integrated flux is proportional to
```math
$$\Large \int_0^z  f(z,\nu) z dz$$
```
has **no divergence** at `z=0` as  `nu>-1` (which is the original paper bound) as nu is in [-0.85, 4.0] (GalSim bounds). 

So, I do not see the need to make the procedure mentionned in GalSim and I do suspect that it introduces the difference detected by pytest. 

Now, despite what I've written above, I will try to implement the GalSim procedure.

jecampagne:
I have opened an issue in the GalSim repo to understand how they manage the case nu<0 as I may found a pb in the normalisation. Let see what Mike tells me (notice that I am not lucky so far when discussing we Mike...)

jecampagne:
Hi, I have implemented a `shoot_neg`(ie. for nu<0) such that 1) we follow the Galsim spirirt of using a linear approx of the profile for low r value and 2) I use ae normalzation which is not impacting the pytest but is math correct. 
I have also use the inevrsion of the cumukative flux along the exponential use-case (ie. lazy_property deco & linear iinterpolation inversion). The whole pytest last 23min now).

jecampagne:
There is still a missing test to assert that nu parameter is in [-0.86, 4.0].

beckermr:
I'll take a look tomorrow!

jax_galsim/spergel.py
```diff
@@ -0,0 +1,470 @@
...
+            I(r) = flux \times \left(2\pi 2^\nu \Gamma(1+\nu) r_0^2\right)^{-1}
+              \times \left(\frac{r}{r_0}\right)^\nu K_\nu\left(\frac{r}{r_0}\right)
+
+    with the following Fourier expression
+        .. math::
+
+            \hat{I}(k) = flux / (1 + (k r_0)^2)^{1+\nu}
+
+    where :math:`r_0` is the ``scale_radius``, and :math: `\nu` mandatory to be in [-0.85,4.0]
+    """,
```
beckermr, line 208: I'm confused by this. Does this convention mean that the spergel profile in galsim vs jax_galsim with the same flux will produce different surface brightness values?

beckermr (changes requested):

jax_galsim/spergel.py
jecampagne: Hi @beckermr , sorry if I have introduced  confusing comments. 
I have followed the same definition as Galsim code, so in principle all the physics paramters are the same. Now, I thought that the Galsim class comment is not so clear: 
1.  most importantly compared to the original Spergel's paper for instance, the parameter r0 in Galsim is the scale_factor and not the half-light-radius as Spergel notation. So it can make people confusing
2. the nu parameter in the JAX-Galsim code is not asserted to be in the [-0.86,4.0] interval (ie. no JAX way to do it till we find a way), so we need to specify this somewhere.
3. I prefer personally to write the complete expression and no let the user discovering it reading the code.

beckermr (changes requested):
I left one suggestion for rephrasing the docs. 

My final comment is that the galsim APIs provide some bessel functions in galsim.bessel. Any bessel functions defined in this file that are provided by galsim in the `galsim.bessel` module should be moved there.

jax_galsim/spergel.py
beckermr: I suggested a change in wording to resolve this.

jecampagne:
I have rename "Knu" (modified Bessel 2nd kind) as "kv"  to follow Galsim notation. Both Gamma & and Knu are moved to `jax_galsim/bessel.py`. I have realised then that Knu was defined & used for Moffat, so I have remove the code to use the new one. I have not moved other z^nu Knu(nu,z) related functions as there are specific to Spergel profile.

Now in the `core`directory there is a "bessel.py" file that I had created in the past. I think this should be removed and the code moved to `jax_galsim/bessel.py`. What do you think?

jax_galsim/bessel.py
```diff
@@ -101,3 +102,15 @@ def si(x):
...
+@jax.jit
+def kv(nu, x):
+    """Modified Bessel 2nd kind"""
+    return tfp.substrates.jax.math.bessel_kve(nu * 1.0, x) / jnp.exp(jnp.abs(x))
+
+
+@jax.jit
+def gamma(x):
+    """Gamma(x)"""
+    return jnp.exp(jax.lax.lgamma(x * 1.0))
```
beckermr: Why do you have `* 1.0` in these functions?

jecampagne: `test_spergel.py` : 

1. _E       TypeError: lgamma does not accept dtype int64. Accepted dtypes are subtypes of floating._
2. using ` galsim.Spergel(nu=0, half_light_radius=3.5, flux=1.e4)` makes a crash in kv

`test_bessel.py` idem for kv: 
_E         TypeError: Found incompatible dtypes, <class 'numpy.float64'> and <class 'numpy.int64'>. Seen so far: [<class 'numpy.float64'>, <class 'numpy.int64'>]_

beckermr (approved):

beckermr:
> Now in the coredirectory there is a "bessel.py" file that I had created in the past. I think this should be removed and the code moved to jax_galsim/bessel.py. What do you think?

No we should not do this. The goal is to match the galsim API exactly. So we only put bessel functions that galsim provides in jax_galsim/bessel.py. Anything extra, we can add to core since this contains private APIs to jax_galsim.

jecampagne:
Who press the `Merge pull request`?

beckermr:
Either! I'll go ahead!

Issue #93, opened 2024-06-03
vendor and modify `_wraps` from `jax._src.numpy.utils` to use for galsim


beckermr:
@ismael-mendoza noticed that `jax` moved their `_wraps` decorator. It turns out this decorator doesn't work for galsim doc strings anyways. We should vendor their utility and modify to work with galsim doc strings.

PR #94, opened 2024-06-03
ENH update to latest JAX version and slight refactor in test suite

ismael-mendoza:
I was thinking it would be good to update to the latest jax version and pin `jax-galsim` to that if possible. Open to discussion on this though.

Changes: 
- `_wraps` to `implements` (still won't work for `galsim` but allows code to run, see #93 )
- pin `tensorflow-probability`, as the newest version does not have the `substrates` packages?

### DIFF
--- a/tests/jax/test_api.py
+++ b/tests/jax/test_api.py
@@ -39,6 +39,7 @@ OK_ERRORS = [
     "One of scale_radius, half_light_radius must be specified",
     "Arguments to Sum must be GSObjects",
     "'ArrayImpl' object has no attribute 'gsparams'",
+    "'jaxlib.xla_extension.ArrayImpl' object has no attribute 'gsparams'"
     "Supplied image must be an Image or file name",
     "Argument to Deconvolution must be a GSObject.",
 ]
--- a/tests/jax/test_api.py
+++ b/tests/jax/test_api.py
@@ -39,7 +39,7 @@ OK_ERRORS = [
     "One of scale_radius, half_light_radius must be specified",
     "Arguments to Sum must be GSObjects",
     "'ArrayImpl' object has no attribute 'gsparams'",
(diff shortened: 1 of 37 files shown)


beckermr (changes requested):
Please do not pin deps. It just puts off the pain of figuring things out and causes issues downstream. Let's go digging and figure out what tensorflow-probability did with the special functions we need from them.

beckermr:
tfp changed their default submodule imports. A direct import of the bessel function we need works fine.

beckermr:
Nest issue here is that pytest version 8 broke the test mangling we do on galsim. I am looking into that now.

tests/jax/test_api.py
```diff
@@ -38,9 +38,10 @@ def test_api_same():
     "One of scale_radius, half_light_radius, or fwhm must be specified",
     "One of scale_radius, half_light_radius must be specified",
     "Arguments to Sum must be GSObjects",
-    "'ArrayImpl' object has no attribute 'gsparams'",
+    "object has no attribute 'gsparams'",
     "Supplied image must be an Image or file name",
     "Argument to Deconvolution must be a GSObject.",
+    "'jaxlib.xla_extension.ArrayImpl' object has no attribute 'lower'",
```
beckermr, line 44:
```suggestion
    "object has no attribute 'lower'",
```

beckermr (approved):

beckermr:
This PR is ready to go. No API changes here, just updates to the tests etc. Shall we merge @ismael-mendoza?

beckermr:
Also, we should squash when merging. The commit history here is a disaster.

ismael-mendoza:
yes I think so! thank you! 

I agree with the squash and merge, I removed the other options from the settings

beckermr:
Ahhhhh we should keep the other options in the settings IMHO. I simply meant for this PR we'd squash.

PR #97, opened 2024-06-08
PERF block reduce in wrapping

### LINKED ISSUE #96: GPU performance degredation with wrapping code
@ismael-mendoza and @aguinot found that the [wrapping code for hermitian images](https://github.com/aguinot/JAX-GalSim/blob/main/jax_galsim/core/wrap_image.py) is causing a big performance degradation on GPUs. We should investigate and fix. 

see this notebook: https://colab.research.google.com/drive/14RQYB4BSPcv-TeHzz-cDkTZa_jqkXfGe?usp=sharing

beckermr:
This PR uses a block-wise reduction algorithm for wrapping images. It should fix the performance issues of our previous double for loop. 

fixes #96

### DIFF
--- a/jax_galsim/core/wrap_image.py
+++ b/jax_galsim/core/wrap_image.py
@@ -4,6 +4,45 @@ import jax
 import jax.numpy as jnp
 
 
+@partial(jax.jit, static_argnames=("nxwrap", "nywrap"))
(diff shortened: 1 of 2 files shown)


ismael-mendoza (approved):
Looks great, thank you both @beckermr and @jecampagne ! I don't have any comments

PR #102, opened 2024-06-21
feat: add more benchmarks and clean up code

beckermr:
This PR adds more benchmarks and cleans up the benchmark code to remove duplication.

PR #101, opened 2024-06-18
perf: faster kval for Lanczos interpolated images via cached splines

beckermr:
This PR implements faster kval computations for Lanczos kernels via a spline interpolant.

PR #103, opened 2024-06-22
fix: ensure jax caches are always cleared for compile benchmarks

beckermr:
This PR fixes a bug in #102 where the jax jit caches were not cleared properly.


beckermr:
Again this PR has no functional changes and is for testing changes in #101. I am going to merge but I can respond to any CR later.

PR #116, opened 2024-09-09
fix: consistent doc string wrapping for angle.py

### LINKED ISSUE #71: consistent wrapping for docstrings
there are different methods and classes that wrap the Galsim API documentation, but some that don't and copy paste instead. 

I think we should change the code to use wrapping whenever possible

beckermr:
This PR ensures we are using consistent doc string wrapping via the `implements` decorator.

closes #71

### DIFF
--- a/jax_galsim/angle.py
+++ b/jax_galsim/angle.py
@@ -30,29 +30,19 @@ class AngleUnit(object):
     valid_names = ["rad", "deg", "hr", "hour", "arcmin", "arcsec"]
 
     def __init__(self, value):
-        """
-        :param value:   The measure of the unit in radians.
-        """
         if isinstance(value, AngleUnit):
             raise TypeError("Cannot construct AngleUnit from another AngleUnit")
         self._value = cast_to_float(value)
 
     @property
+    @implements(_galsim.AngleUnit.value)
     def value(self):
-        """A read-only attribute giving the measure of the AngleUnit in radians."""
         return self._value
 
     def __rmul__(self, theta):
-        """float * AngleUnit returns an Angle"""
         return Angle(theta, self)
 
     def __div__(self, unit):
-        """AngleUnit / AngleUnit returns a float giving the relative scaling.
-
-        Note: At least to within machine precision, it is the case that
-
-            (x * angle_unit1) / angle_unit2 == x * (angle_unit1 / angle_unit2)
-        """
         if not isinstance(unit, AngleUnit):
             raise TypeError("Cannot divide AngleUnit by %s" % unit)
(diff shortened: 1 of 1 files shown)


jax_galsim/angle.py
```diff
@@ -30,16 +30,13 @@ class AngleUnit(object):
     valid_names = ["rad", "deg", "hr", "hour", "arcmin", "arcsec"]
 
     def __init__(self, value):
-        """
-        :param value:   The measure of the unit in radians.
-        """
```
beckermr, line 35: The init should not have a doc string per normal python conventions.

jax_galsim/gsobject.py
```diff
@@ -213,15 +207,8 @@ def xValue(self, *args, **kwargs):
         pos = parse_pos_args(args, kwargs, "x", "y")
         return self._xValue(pos)
 
+    @implements(_galsim.GSObject._xValue)
```
beckermr, line 210: There are a few formally private but clearly public-facing methods like this one that I went ahead and wrapped.

jax_galsim/interpolatedimage.py
```diff
@@ -232,16 +232,19 @@ def _stepk(self):
             return super()._stepk
 
     @property
+    @implements(_galsim.interpolatedimage.InterpolatedImage.x_interpolant)
     def x_interpolant(self):
         """The real-space `Interpolant` for this profile."""
```
ismael-mendoza, line 237:
```suggestion
```

meant to remove these?

jax_galsim/interpolatedimage.py
```diff
@@ -232,16 +232,19 @@ def _stepk(self):
...
         return self._original._x_interpolant
 
     @property
+    @implements(_galsim.interpolatedimage.InterpolatedImage.k_interpolant)
     def k_interpolant(self):
         """The Fourier-space `Interpolant` for this profile."""
```
ismael-mendoza, line 243:
```suggestion
```

jax_galsim/interpolatedimage.py
```diff
@@ -232,16 +232,19 @@ def _stepk(self):
...
         return self._original._k_interpolant
 
     @property
+    @implements(_galsim.interpolatedimage.InterpolatedImage.image)
     def image(self):
         """The underlying `Image` being interpolated."""
```
ismael-mendoza, line 249:
```suggestion
```

ismael-mendoza (approved):
Just a few small changes, otherwise looks great! Thanks, this made the codebase much smaller

Issue #123, opened 2024-12-10
find optimizations when drawing spergel profiles


beckermr:
I've heard reports that the spergel profiles are much slower than galsim. 

To Do:

- [ ] profile to find optimizations
- [x] add benchmark comparison to galsim
- [x] implement those optimizations

ismael-mendoza:
quick note, when I checked it was a sum of Exponential and Spergel profile (nu=-0.6) -- might be worth taking a look at the timing of each of these separately, but my guess is that it's very likely coming from the Spergel...

beckermr:
We've determined that this is from the conversion of half light radius to scale radius. So we're going to focus our optimization on those bits of the code.

ismael-mendoza:
xref: #205

PR #125, opened 2024-12-19
feat: add benchmarks for spergel profile

beckermr:
This PR adds a benchmark for the Spergel profile. I can indeed confirm it is a few times slower than galsim on the CPU.

xref: #123


ismael-mendoza (approved):
Thanks @beckermr this looks good to me! Perhaps consider adding a benchmark with the HLR liked we talked bout?

PR #141, opened 2025-01-30
perf: try using fori_loop but unrolling everything

beckermr:
I am seeing if this has the same performance as the manual loops.


jax_galsim/core/utils.py
```diff
@@ -254,7 +254,7 @@ def _func(i, args):
     flow = func(low)
     fhigh = func(high)
     args = (func, low, flow, high, fhigh)
-    return jax.lax.fori_loop(0, niter, _func, args)[-2]
+    return jax.lax.fori_loop(0, niter, _func, args, unroll=True)[-2]
```
beckermr, line 257:
```suggestion
    return jax.lax.fori_loop(0, niter, _func, args)[-2]
```

PR #163, opened 2026-02-06
add to_galsim for Image class, `demo1.py`, & `demo2.py`

b-remy:
This PR adds a new HSM module that calls the standard GalSim implementation. The code is almost identical to `hsm.py` from GalSim, with minimal adaptations to accept JAX-GalSim objects as inputs.

I also added `Image`, `Bounds`, and `Position` methods that return the underlying GalSim C++ objects, which are required for calls into the `_galsim` C++ layer.

In addition, this PR introduces `demo1.py` and `demo2.py`, which are direct copies of the corresponding GalSim demos, modified only to import `jax_galsim`.

The logs produced by GalSim and JAX-GalSim differ slightly. This may be related to differences in how images are drawn in JAX-GalSim rather than the HSM computation itself. I’d be curious to hear your thoughts on this @beckermr.

## `demo1.py`
GalSim
```
Starting demo script 1 using:
    - circular Gaussian galaxy (flux = 1.0e+05, sigma = 2.0),
    - circular Gaussian PSF (sigma = 1.0),
    - pixel scale = 0.20,
    - Gaussian noise (sigma = 30.00).
Wrote image to 'output/demo1.fits'
HSM reports that the image has observed shape and size:
    e1 = -0.025, e2 = -0.009, sigma = 11.060 (pixels)
Expected values in the limit that pixel response and noise are negligible:
    e1 = 0.000, e2 = 0.000, sigma = 11.180
```
JAX-GalSim
```
Starting demo script 1 using:
    - circular Gaussian galaxy (flux = 1.0e+05, sigma = 2.0),
    - circular Gaussian PSF (sigma = 1.0),
    - pixel scale = 0.20,
    - Gaussian noise (sigma = 30.00).
Wrote image to 'output/demo1.fits'
HSM reports that the image has observed shape and size:
    e1 = 0.008, e2 = 0.043, sigma = 11.010 (pixels)
Expected values in the limit that pixel response and noise are negligible:
    e1 = 0.000, e2 = 0.000, sigma = 11.180
```

## `demo2.py`
GalSim
```
Starting demo script 2 using:
    - sheared (0.10,0.20) exponential galaxy (flux = 1.0e+05, scale radius = 2.70),
    - circular Moffat PSF (beta = 5.0, re = 1.00),
    - pixel scale = 0.20,
    - Poisson noise (sky level = 2.5e+03).
Wrote image to 'output/demo2.fits'
Wrote effective PSF image to 'output/demo2_epsf.fits'
HSM reports that the image has observed shape and size:
    e1 = 0.161, e2 = 0.343, sigma = 16.714 (pixels)
When carrying out Regaussianization PSF correction, HSM reports distortions
    e1, e2 = 0.176, 0.376
Expected values in the limit that noise and non-Gaussianity are negligible:
    g1, g2 = 0.190, 0.381
```
JAX-GalSim
```
Starting demo script 2 using:
    - sheared (0.10,0.20) exponential galaxy (flux = 1.0e+05, scale radius = 2.70),
    - circular Moffat PSF (beta = 5.0, re = 1.00),
    - pixel scale = 0.20,
    - Poisson noise (sky level = 2.5e+03).
/home/ben/Documents/repos/probabilistic-shear/.venv/lib/python3.10/site-packages/jax/_src/numpy/lax_numpy.py:188: UserWarning: Explicitly requested dtype float64 requested in asarray is not available, and will be truncated to dtype float32. To enable more dtypes, set the jax_enable_x64 configuration option or the JAX_ENABLE_X64 shell environment variable. See https://github.com/jax-ml/jax#current-gotchas for more.
  return asarray(x, dtype=self.dtype)
Wrote image to 'output/demo2.fits'
Wrote effective PSF image to 'output/demo2_epsf.fits'
HSM reports that the image has observed shape and size:
    e1 = 0.172, e2 = 0.352, sigma = 17.127 (pixels)
When carrying out Regaussianization PSF correction, HSM reports distortions
    e1, e2 = 0.188, 0.385
Expected values in the limit that noise and non-Gaussianity are negligible:
    g1, g2 = 0.190, 0.381
```

### DIFF
--- a/examples/demo2.py
+++ b/examples/demo2.py
@@ -1,0 +1,175 @@
+# Copyright (c) 2012-2026 by the GalSim developers team on GitHub
+# https://github.com/GalSim-developers
+#
+# This file is part of GalSim: The modular galaxy image simulation toolkit.
+# https://github.com/GalSim-developers/GalSim
+#
+# GalSim is free software: redistribution and use in source and binary forms,
+# with or without modification, are permitted provided that the following
+# conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice, this
+#    list of conditions, and the disclaimer given in the accompanying LICENSE
+#    file.
+# 2. Redistributions in binary form must reproduce the above copyright notice,
+#    this list of conditions, and the disclaimer given in the documentation
+#    and/or other materials provided with the distribution.
+#
+"""
+Demo #2
+
+The second script in our tutorial about using JAX-GalSim in python scripts: examples/demo*.py.
+(This file is designed to be viewed in a window 100 characters wide.)
+
+This script is a bit more sophisticated, but still pretty basic.  We're still only making
+a single image, but now the galaxy has an exponential radial profile and is sheared.
+The PSF is a circular Moffat profile.  The noise is drawn from a Poisson distribution
+using the flux from both the object and a background sky level to determine the
+variance in each pixel.
+
+New features introduced in this demo:
+
+- obj = jax_galsim.Exponential(flux, scale_radius)
+- obj = jax_galsim.Moffat(beta, flux, half_light_radius)
+- obj = obj.shear(g1, g2)  -- with explanation of other ways to specify shear
+- rng = jax_galsim.BaseDeviate(seed)
+- noise = jax_galsim.PoissonNoise(rng, sky_level)
+- galsim.hsm.EstimateShear(image, image_epsf)
+"""
+
+import logging
+import os
+import sys
+
+import galsim
+
+import jax_galsim
+
+
+def main(argv):
+    """
+    A little bit more sophisticated, but still pretty basic:
+      - Use a sheared, exponential profile for the galaxy.
+      - Convolve it by a circular Moffat PSF.
+      - Add Poisson noise to the image.
+    """
+    # In non-script code, use getLogger(__name__) at module scope instead.
+    logging.basicConfig(format="%(message)s", level=logging.INFO, stream=sys.stdout)
+    logger = logging.getLogger("demo2")
+
+    gal_flux = 1.0e5  # counts
+    gal_r0 = 2.7  # arcsec
+    g1 = 0.1  #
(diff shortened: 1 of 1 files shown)


beckermr (changes requested):
So I am looking at this PR more, and I think a simpler approach is warranted. 

Right now we've made modifications to the jax-galsim images to monkeypatch in galsim stuff in the various places where galsim normally has a C++ object. This is painful and is causing random extra bits of code to splayed all over the repo because now sometimes a pure galsim object comes by when only jax-galsim items were expected before.

We need to maintain a strict boundary between the two code bases.


So, instead of trying to directly pass jax-galsim stuff to galsim, we should instead expand on `from_galsim` API in other parts of the repo by adding a `to_galsim` method to the image classes. This method would convert the image to a pure galsim image. Then we can feed this directly to HSM. 

For the returned data, we should not bother wrapping it in our own class or anything. We should return directly the galsim ShapeData class.

beckermr:
Also if you have any debugging info on the nans for demo2, that'd be helpful.

jax_galsim/image.py
```diff
@@ -1094,15 +1094,40 @@ def tree_unflatten(cls, aux_data, children):
...
+    @implements(
+        _galsim.Image.FindAdaptiveMom,
+        lax_description=(
+            "This method converts the current `jax_galsim.Image` to a native "
+            "`galsim.Image` and delegates the computation to "
+            "`galsim.hsm.FindAdaptiveMom`. The returned object is GalSim's "
+            "`ShapeData`."
+        ),
+    )
+    def FindAdaptiveMom(self, *args, **kwargs):
```
beckermr: I think we need to ensure we convert any arguments or keyword arguments as well. Those could be jax_galsim objects.

b-remy:
Good to know, thank you!

b-remy:
Thanks @beckermr ! And thanks to #181 I can also add `demo2.py` to this PR

PR #164, opened 2026-02-07
Minimal JAX version bump

EiffL:
This PR aims for minimal edits allowing for jax version update. With this update, we can reach the latest jax 0.8 version, quadax prevents us from going to jax v0.9.

TFP is apparentently no longer releasing new versions, but tfp-nightly continues to be maintained for new jax releases. We should still ultimately remove TFP because it's only used for this very speicific Bessel function, and it prevents installs on python 3.13, but for now this would be an easy fix and we can then compare the code speed before and after changing the bessel function implementation on same version of jax.

I left the tfp-nighlty version unconstrained, it should automatically adjust to the jax version, which in turn will be the latest version permitted by quadrax.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,7 @@ readme = "README.md"
 dependencies = [
     "numpy >=1.18.0",
     "galsim >=2.3.0",
-    "jax >=0.4.0",
+    "jax >=0.8.0",
     "astropy >=2.0",
     "tfp-nightly",
     "quadax",


pyproject.toml
```diff
@@ -16,10 +16,9 @@ readme = "README.md"
 dependencies = [
     "numpy >=1.18.0",
     "galsim >=2.3.0",
-    "jax <0.7.0",
-    "jaxlib",
+    "jax >=0.4.0",
```
beckermr:
```suggestion
    "jax >=0.8.0",
```

PR #165, opened 2026-02-07
Add git-based dynamic versioning with setuptools-scm

EiffL:
## Summary

- Replace hardcoded `version = "0.0.1rc1"` with `setuptools-scm`, which derives versions automatically from git tags
- Expose `jax_galsim.__version__` at runtime via a generated `_version.py`
- Update CI to fetch full git history (`fetch-depth: 0`) so setuptools-scm can resolve the version, and use `submodules: recursive` in the checkout step instead of a manual `git submodule update`

## How it works

`setuptools-scm` reads git tags at build/install time:
- **At a tag** like `v0.0.1rc1` → version is `0.0.1rc1`
- **N commits past a tag** → version is `0.0.1rc1.devN+g<hash>`
- **No tag found** → falls back to `0.0.1rc1.dev0` (configured in `pyproject.toml`)

## Action needed after merge

To anchor the versioning, a maintainer should create an initial SemVer tag on the merge commit:

```bash
git tag v0.0.1rc1
git push origin v0.0.1rc1
```

We recommend **SemVer** (`MAJOR.MINOR.PATCH`) over CalVer for this project because:
- As a library, downstream code depends on the API — SemVer communicates compatibility (breaking vs non-breaking changes)
- Scientific computing users need clear signals about whether an upgrade is safe
- The `0.x.y` convention explicitly signals "pre-stable, API may change"

**Also, I don't think setuptools-scm works with CalVer**

### DIFF
--- a/.github/workflows/python_package.yaml
+++ b/.github/workflows/python_package.yaml
@@ -27,6 +27,7 @@ jobs:
         with:
           fetch-depth: 0
           submodules: recursive
+          fetch-tags: true
 
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v6


.github/workflows/python_package.yaml
```diff
@@ -24,6 +24,9 @@ jobs:
 
     steps:
       - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+          submodules: recursive
```
beckermr:
```suggestion
          submodules: recursive
          fetch-tags: true
```

Sometimes I need this for SCM and sometimes I do not and I do not know when or why. Could be good to add it.

beckermr:
See calver-by-date here: https://setuptools-scm.readthedocs.io/en/latest/extending/#available-implementations

.github/workflows/python_package.yaml
EiffL: yes I think the 'fetch-depth: 0' is supposed to fetch the entire history including tags. Maybe not needed if 'fetch-tags: true'
I don't think there is any harm in having both

beckermr:
If we switch to calver we may want to change the default version. We can do that in a future PR.

PR #171, opened 2026-02-08
perf: make test suite snappier and clean up some testing things

### LINKED ISSUE #166: adopt pytest-xdist for faster tests
We should try this one.

beckermr:
This PR makes the test suite snappier by:

- using pytest-split to split the tests over four jobs
- removing extra benchmarks

I've also fixed a few deprecations, bugs, and minor oddities in the tests to make life easier. This includes

- updating the pre-commit hooks to use the correct hook for ruff-check
- changed codspeed mode to simulation to match latest API
- changing imports in the tests to allow them to be run without import errors due to not having galsim's test utils module
- adding a call to make the `output` directory in our text fixture if it does not exist
- removed premade output directory in favor of making it on the fly

closes #166

### DIFF
--- a/tests/output/README.md
+++ b/tests/output/README.md
@@ -1,1 +1,0 @@
-Folder for storing test outputs


tests/conftest.py
```diff
@@ -18,6 +18,9 @@
 # Identify the path to this current file
 test_directory = os.path.dirname(os.path.abspath(__file__))
 
+# make the output directory for the test output
+os.makedirs(os.path.join("..", "output"), exist_ok=True)
```
beckermr:
```suggestion
os.makedirs("output", exist_ok=True)
```

Issue #161, opened 2026-02-06
moffat profile drawing is slow


ismael-mendoza:
need to dig into this more, some of us have a suspicion it has to do with the `bisect_for_root` usage in this function: 

https://github.com/GalSim-developers/JAX-GalSim/blob/f1dcdd7fb5ceb1dfe24f9b646357d3855bc19931/jax_galsim/moffat.py#L290

beckermr:
Yeah this is in principle a 4d function of the profile HLR, profle beta, profile truncation raidus, and the galsim numerical threshold.

ismael-mendoza:
yes @beckermr I can take a look

ismael-mendoza:
I find a very significant speedup compared to the old version of the code. The test is very simple just drawing a Gaussian galaxy with either a Gaussian or Moffat PSF with fixed parameters (jitted in a NERSC A100 GPU). The slen was set to 63 and the fft_size to 256.

```python
def draw_gaussian_gaussian_psf(
    *,
    f: float,
    hlr: float,
    e1: float,
    e2: float,
    x: float,  # pixels
    y: float,
    slen: int,
    fft_size: int,  # rule of thumb: at least 4 times `slen`
    psf_fwhm: float = 0.8,
    pixel_scale: float = 0.2,
):
    gsparams = GSParams(minimum_fft_size=fft_size, maximum_fft_size=fft_size)

    gal = xgalsim.Gaussian(flux=f, half_light_radius=hlr)
    gal = gal.shear(g1=e1, g2=e2)

    psf = xgalsim.Gaussian(flux=1.0, fwhm=0.8)
    gal_conv = xgalsim.Convolve([gal, psf]).withGSParams(gsparams)
    image = gal_conv.drawImage(nx=slen, ny=slen, scale=pixel_scale, offset=(x, y))
    return image.array

def draw_gaussian_moffat_psf(
    *,
    f: float,
    hlr: float,
    e1: float,
    e2: float,
    x: float,  # pixels
    y: float,
    slen: int,
    fft_size: int,  # rule of thumb: at least 4 times `slen`
    psf_fwhm: float = 0.8,
    pixel_scale: float = 0.2,
):
    gsparams = GSParams(minimum_fft_size=fft_size, maximum_fft_size=fft_size)

    gal = xgalsim.Gaussian(flux=f, half_light_radius=hlr)
    gal = gal.shear(g1=e1, g2=e2)

[...]

PR #180, opened 2026-02-08
update bessel algo

### LINKED ISSUE #179: Moffat profile drawing returns NaN (`demo2.py`)
Running GalSim's `demo2.py` with JAX-GalSim fails and it seems to come from the Moffat profile. Here's a minimal script showing the error

```python
import jax_galsim

gal_flux = 1.e5    # counts
gal_r0 = 2.7       # arcsec
g1 = 0.1           #
g2 = 0.2           #
psf_beta = 5       #
psf_re = 1.0       # arcsec
pixel_scale = 0.2  # arcsec / pixel

# Define the galaxy profile.
gal = jax_galsim.Exponential(flux=gal_flux, scale_radius=gal_r0)

# Shear the galaxy by some value.
gal = gal.shear(g1=g1, g2=g2)

# Define the PSF profile.
psf = jax_galsim.Moffat(beta=psf_beta, flux=1., half_light_radius=psf_re)

# Final profile is the convolution of these.
final = jax_galsim.Convolve([gal, psf])

print(gal.drawImage(scale=pixel_scale).array)
"""
[[3.8022862e-04 3.8958775e-04 4.0335540e-04 ... 6.2590830e-06
  7.4280219e-06 6.6112520e-06]
 [4.0020858e-04 4.1460019e-04 4.2499771e-04 ... 7.5581743e-06
  6.7426722e-06 7.9196307e-06]
 [4.2534867e-04 4.3639768e-04 4.5194468e-04 ... 6.8268273e-06
  7.9949132e-06 7.2239072e-06]
 ...
 [7.2855555e-06 8.0750069e-06 6.9304629e-06 ... 4.5205365e-04
  4.3652899e-04 4.2551477e-04]
 [7.2642902e-06 6.0641905e-06 6.8564354e-06 ... 4.2461159e-04
  4.1419003e-04 3.9978177e-04]
 [6.7501096e-06 7.5687421e-06 6.3996608e-06 ... 4.0379266e-04
  3.9000134e-04 3.8062737e-04]]
"""

print(psf.drawImage(scale=pixel_scale).array)
"""
[[nan nan nan ... nan nan nan]
 [nan nan nan ... nan nan nan]
 [nan nan nan ... nan nan nan]
 ...
 [nan nan nan ... nan nan nan]
 [nan nan nan ... nan nan nan]
 [nan nan nan ... nan nan nan]]
"""

print(final.drawImage(scale=pixel_scale).array)
"""
[[nan nan nan ... nan nan nan]
 [nan nan nan ... nan nan nan]
 [nan nan nan ... nan nan nan]
 ...
 [nan nan nan ... nan nan nan]
 [nan nan nan ... nan nan nan]
 [nan nan nan ... nan nan nan]]
"""
```

EiffL:
rewrite of bessel kv wth fixed length iterations, and as proper jax primitives.

this implementation was protiyped here: https://gist.github.com/EiffL/95e00c160dc42a58cb44b0bf8aab83ea

### DIFF
--- a/jax_galsim/bessel.py
+++ b/jax_galsim/bessel.py
@@ -207,7 +207,7 @@ _ASYMPTOTIC_OLVER_COEFFICIENTS = [
 
 def _sqrt1px2(x):
     """Numerically stable sqrt(1 + x^2)."""
-    eps = _const(x, jnp.finfo(jnp.float64).eps)
+    eps = _const(x, jnp.finfo(x.dtype).eps)
     return jnp.where(
         jnp.abs(x) * jnp.sqrt(eps) <= _const(x, 1.0),
         jnp.exp(_const(x, 0.5) * jnp.log1p(x * x)),
@@ -259,7 +259,7 @@ def _temme_series_kve(v, z):
     Uses fori_loop with fixed 15 iterations (empirically, max needed is 12
     for f64 across the valid domain).
     """
-    tol = _const(z, jnp.finfo(jnp.float64).eps)
+    tol = _const(z, jnp.finfo(z.dtype).eps)
 
     coeff1, coeff2, gamma1pv_inv, gamma1mv_inv = _evaluate_temme_coeffs(v)
 
@@ -324,7 +324,7 @@ def _continued_fraction_kve(v, z):
     Uses fori_loop with fixed 80 iterations (empirically, max needed is 77
     for f64 at z~2).
     """
-    tol = _const(z, jnp.finfo(jnp.float64).eps)
+    tol = _const(z, jnp.finfo(z.dtype).eps)
 
     initial_numerator = v * v - _const(v, 0.25)
     initial_denominator = _const(z, 2.0) * (z + _const(z, 1.0))
--- a/jax_galsim/bessel.py
+++ b/jax_galsim/bessel.py
@@ -1,6 +1,18 @@
 import galsim as _galsim
 import jax
 import jax.numpy as jnp
+import numpy as np
+from jax._src import core, dtypes
+from jax._src.interpreters import ad, mlir
+from jax._src.lax.lax import (
+    _const,
+    _float,
+    broadcast_in_dim,
+    broadcast_shapes,
+    convert_element_type,
+    standard_naryop,
+)
+from jax._src.lax.special import evaluate_chebyshev_polynomial
 
 from jax_galsim.core.utils import implements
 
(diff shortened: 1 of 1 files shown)


EiffL:
## How this PR addresses #161

Issue #161 identifies that Moffat profile drawing is slow, with gradients being **40x more expensive** than the forward pass. The detailed analysis by @EiffL pinpoints several contributing factors in the `kv()` Bessel function implementation (listed as "Fix 4" in the issue). This PR rewrites `kv()` as a proper JAX primitive, which addresses several of those root causes:

### 1. No more unnecessary forward-pass work for gradients

The old implementation used `@jax.custom_vjp`, which **unconditionally computes 3 Bessel evaluations** (K_v, K_{v-1}, K_{v+1}) in every forward pass and stores them as residuals for the backward pass — even when no gradient is needed (e.g., when evaluating the PSF in a convolution). From the issue:

> The `kv()` Bessel function (`bessel.py:550-554`) always computes 3 Bessel evaluations in the forward pass to store residuals for the backward.

The new implementation uses `ad.defjvp` on the primitive instead. The JVP rule is only invoked during differentiation — the forward pass computes only the single K_v evaluation it actually needs. The K_{v-1} and K_{v+1} terms are computed lazily, only when JAX's AD system actually requests the derivative.

### 2. Fixed-size `fori_loop` replaces `while_loop` → smaller HLO

The old code used `jax.lax.while_loop` for convergence-based iteration in the Temme series and Steed's continued fraction algorithms. `while_loop` generates larger, harder-to-optimize HLO because XLA cannot predict the iteration count at compile time.

The new code uses `jax.lax.fori_loop` with empirically-determined fixed iteration counts (15 for Temme series, 80 for Steed CF), using `jnp.where(converged, old_val, new_val)` to no-op after convergence. This lets XLA:
- Know the exact loop trip count at compile time
- Potentially unroll or pipeline the loops
- Generate more compact HLO graphs

This is relevant to the observation in the issue that "taking gradients of the moffat and exponential convolution creates a very big HLO."

### 3. Native element-wise operation replaces `vmap` over scalar core

The old implementation defined a scalar `_kv_scalar` function and used `jax.vmap` to vectorize it over array inputs. The new implementation operates element-wise natively using `_up_and_broadcast` (the same pattern used by `igamma`, `betainc`, and other special functions in `jax._src.lax.special`). This removes one layer of `vmap` tracing overhead from the evaluation stack — relevant to @beckermr's point that "XLA is not smart enough to do all of its optimizations with vmap."

### 4. Proper JAX primitive integration

Registering `kv` as a JAX primitive via `standard_naryop` + `mlir.register_lowering` + `ad.defjvp` means it integrates with JAX's compiler infrastructure the same way built-in special functions do, rather than being an opaque Python-level overlay. This gives XLA maximum visibility into the operation for optimization.

### What this doesn't fix

This PR addresses Fix 4 from the issue analysis. The other identified improvements — stopping gradient flow through non-parameter PSF objects (Fix 2) and vectorizing k-space evaluations in `draw_by_kValue` (Fix 3) — are orthogonal and would need separate PRs.

### Accuracy & performance

- Max relative error vs scipy.special.kv: ~1.2e-13 across a wide (v, x) grid
- All 28 existing gradient tests pass
- Gradient computation uses the analytical recurrence relation: ∂K_ν/∂x = -½(K_{ν-1}(x) + K_{ν+1}(x))

EiffL:
I haven't yet confirmed this PR is helping the original problem of inefficient moffats and gradients, so I'll keep it as a draft for now. Should be able to run that test tonight

PR #194, opened 2026-02-12
fix: nan in derivs wrt params for Moffat

beckermr:
This PR 

- fixes NaN derivatives for the Moffat profile wrt its parameters
- adds tests for the derivs

I separated this fix from the performance fixes in #188 to keep things clean.


beckermr:
@EiffL can you look this one over?

PR #220, opened 2026-04-16
test: update tests with latest commits from main

beckermr:
This PR added some changes:

- Some of the numeric operations now round to int types like galsim does.
- More of the code uses the jax `.at` syntax to enable in-place operations when things are inside JIT.
- Arrays read from numpy are converted to native byteorder before they are used in JAX. This seems to be needed with jax version 0.10 and later.

### DIFF
--- a/.github/workflows/python_package.yaml
+++ b/.github/workflows/python_package.yaml
@@ -75,10 +75,17 @@ jobs:
             --retries 1
 
       - name: Test with pytest in float32
-        if: ${{ matrix.group == '1' }}
         run: |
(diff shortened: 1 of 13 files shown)


.github/workflows/python_package.yaml
```diff
@@ -64,7 +64,7 @@ jobs:
       - name: Test with pytest
         run: |
           pytest \
-            -vv \
+            -vvx \
```
beckermr:
```suggestion
            -vv \
```

Issue #189, opened 2026-02-10
some numeric thresholds are calculated differently from galsim


ismael-mendoza:
@beckermr pointed out that some numeric thresholds related to drawing images like `maxk` are not computed in the same way between jax galsim and galsim. As I understand it, this doesn't lead to significantly biased results and relates to the fact that some of the way these algorithms were implemented in galsim are not very jax friendly. The problem is that the implementation in jax galsim for these algorithms leads to different outputs  than galsim everything else equal which might surprise the user. 

We discussed in the meeting two options: 

(1) Document these caveats so the users know about them
(2) Try to find new implementations in jax that match the galsim output

I think we were leaning towards (1). 

@beckermr do you have any particular example of this? Does the above accurately describe the issue you had in mind?

ismael-mendoza:
For InterpolatedImage _stepk/_maxk, see these tests for instance

- https://github.com/GalSim-developers/JAX-GalSim/blob/d9f01555a944f72ad73bc43efcdc9b582d8004c2/tests/jax/test_interpolatedimage_utils.py#L118

Also useful the corresponding functions in GalSim ( in the c++ layer)
- [calculateStepK](https://github.com/GalSim-developers/GalSim/blob/ee177aa50424dbbc5cec4a71cadd6297109c2aa1/src/SBInterpolatedImage.cpp#L1146)
- [calculateMaxK](https://github.com/GalSim-developers/GalSim/blob/ee177aa50424dbbc5cec4a71cadd6297109c2aa1/src/SBInterpolatedImage.cpp#L1117)

beckermr:
xref: https://github.com/GalSim-developers/JAX-GalSim/pull/224

PR #221, opened 2026-04-23
fix: make stepk match galsim

beckermr:
This PR gets stepk for interpolated images to match galsim exactly. 🎉 

xref: #189

### DIFF
--- a/jax_galsim/interpolatedimage.py
+++ b/jax_galsim/interpolatedimage.py
@@ -378,6 +378,7 @@ class _InterpolatedImageImpl(GSObject):
     WCS handling bits. For those, we inherit from the Transform class so that
     we can reuse its methods.
     """
+
     _cache_noise_pad = {}
 
     _has_hard_edges = False
--- a/jax_galsim/interpolatedimage.py
+++ b/jax_galsim/interpolatedimage.py
@@ -224,7 +224,7 @@ class InterpolatedImage(Transformation, metaclass=DirMeta):
         else:
             # galsim uses a different way to handfle the WCS effects on maxk
             # for interpolated images. IDK why. - MRB
-            return self._original.stepk / self._original._wcs._minScale()
+            return self._original.maxk / self._original._wcs._maxScale()
 
     @property
     def _stepk(self):
--- a/jax_galsim/interpolatedimage.py
+++ b/jax_galsim/interpolatedimage.py
@@ -19,7 +19,6 @@ from jax.tree_util import register_pytree_node_class
 from jax_galsim import fits
 from jax_galsim.bounds import BoundsI
 from jax_galsim.core.utils import (
-    compute_major_minor_from_jacobian,
     ensure_hashable,
     implements,
 )
@@ -685,16 +684,14 @@ class _InterpolatedImageImpl(GSObject):
     @property
     def _maxk(self):
         if self._jax_aux_data["_force_maxk"]:
-            _, minor = compute_major_minor_from_jacobian(self._jac_arr.reshape((2, 2)))
-            return self._jax_aux_data["_force_maxk"] * minor
+            return self._jax_aux_data["_force_maxk"]
         else:
             return self._getMaxK(self._jax_aux_data["calculate_maxk"])
 
     @property
     def _stepk(self):
         if self._jax_aux_data["_force_stepk"]:
-            major, _ = compute_major_minor_from_jacobian(self._jac_arr.reshape((2, 2)))
-            return self._jax_aux_data["_force_stepk"] * major
+            return self._jax_aux_data["_force_stepk"]
         else:
             return self._getStepK(self._jax_aux_data["calculate_stepk"])
 
--- a/jax_galsim/interpolatedimage.py
+++ b/jax_galsim/interpolatedimage.py
@@ -688,7 +688,7 @@ class _InterpolatedImageImpl(GSObject):
         else:
             # the factor of 1.1 here is a fudge to make jax_galsim a bit more
             # conservative when computing maxk
-            return self._getMaxK(self._jax_aux_data["calculate_maxk"]) * 1.1
+            return self._getMaxK(self._jax_aux_data["calculate_maxk"]) * 1.03
 
     @property
     def _stepk(self):
--- a/jax_galsim/interpolatedimage.py
+++ b/jax_galsim/interpolatedimage.py
@@ -1239,6 +1239,6 @@ def _find_maxk(kim, max_maxk, thresh):
         # since compared to galsim, it does NOT require 5 rows
         # of pixels in a row below the threshold.
         # thus we add some pixels here to ensure the galsim tests pass.
-        _inner_comp_find_maxk(kim.array, thresh, kx, ky) + 20 * kim.scale,
+        _inner_comp_find_maxk(kim.array, thresh, kx, ky) + 1 * kim.scale,
         max_maxk,
     )
--- a/jax_galsim/interpolatedimage.py
+++ b/jax_galsim/interpolatedimage.py
@@ -219,7 +219,7 @@ class InterpolatedImage(Transformation, metaclass=DirMeta):
         if self._jax_aux_data["_force_maxk"] > 0:
             return self._jax_aux_data["_force_maxk"]
         else:
-            # galsim uses a different way to handfle the WCS effects on maxk
+            # galsim uses a different way to handle the WCS effects on maxk
             # for interpolated images. IDK why. - MRB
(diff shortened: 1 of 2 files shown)


jax_galsim/interpolatedimage.py
```diff
@@ -1175,7 +1175,7 @@ def _flux_frac(a, x, y, cenx, ceny):
     dy = jnp.reshape(dy, (a.shape[0], a.shape[1], 1))
     d = jnp.arange(a.shape[0])
     d = jnp.reshape(d, (1, 1, -1))
-    msk = (jnp.abs(dx) <= d) & (jnp.abs(dx) <= d)
+    msk = (jnp.abs(dx) <= d) & (jnp.abs(dy) <= d)
```
beckermr, line 1166: This was a legit bug. :(

jax_galsim/interpolatedimage.py
```diff
@@ -687,13 +686,15 @@ def _maxk(self):
...
+            # conservative when computing maxk
+            return self._getMaxK(self._jax_aux_data["calculate_maxk"]) * 1.1
 
     @property
     def _stepk(self):
         if self._jax_aux_data["_force_stepk"]:
-            _, minor = compute_major_minor_from_jacobian(self._jac_arr.reshape((2, 2)))
-            return self._jax_aux_data["_force_stepk"] * minor
+            major, _ = compute_major_minor_from_jacobian(self._jac_arr.reshape((2, 2)))
+            return self._jax_aux_data["_force_stepk"] * major
```
beckermr: This was also a bug. 🤦

jax_galsim/interpolatedimage.py
```diff
@@ -1193,28 +1192,20 @@ def _calculate_size_containing_flux(image, thresh):
     cenx, ceny = image.center.x, image.center.y
     x, y = image.get_pixel_centers()
     fluxes = _flux_frac(image.array, x, y, cenx, ceny)
-    msk = fluxes >= -jnp.inf
-    fluxes = jnp.where(msk, fluxes, jnp.max(fluxes))
+    # msk = fluxes >= -jnp.inf
+    # fluxes = jnp.where(msk, fluxes, jnp.max(fluxes))
     d = jnp.arange(image.array.shape[0]) + 1.0
```
ismael-mendoza: curious if this assumes the image is a square? GalSim does something with the minimum in the x and y direction to set this maximum distance no?

beckermr: Yes we can compute a min across the axis sizes.

beckermr:
Test failures here are due to maxk not matching.

jax_galsim/interpolatedimage.py
ismael-mendoza: could you explain if you don't mind? I couldn't find the logic in GalSim multiplying times the jacobian major/minor anywhere

beckermr: It is in the transform class and this bit of code undoes what the transform class will do later.

beckermr:
Let's go ahead with this PR for now. I still want to look into maxk more, but this PR leaves the code in a pretty good working state.

beckermr:
@ismael-mendoza this is ready for review!

ismael-mendoza:
thanks! I can take a look today

jax_galsim/interpolatedimage.py
ismael-mendoza: not sure whether to leave this for another PR... but do we need `self._original._wcs._maxScale()` here like for stepk we have `self._original._wcs._minScale()`?

And the Transform `super()` there applies `minor`, not `major`...

I see that GalSim does it here: https://github.com/GalSim-developers/GalSim/blob/ee177aa50424dbbc5cec4a71cadd6297109c2aa1/galsim/interpolatedimage.py#L626, but I don't think we do?

ismael-mendoza: this might depend on the changes from the above comment, but do we need `minor` in this line to remove the `minor` from the `Transform` above?

jax_galsim/interpolatedimage.py
```diff
@@ -1193,28 +1191,18 @@ def _calculate_size_containing_flux(image, thresh):
...
-    # below we use a linear interpolation table to find the maximum size
-    # in pixels that contains a given flux (called thresh here)
-    # expfac controls how much we oversample the interpolation table
-    # in order to return a more accurate result
-    # we have it hard coded at 4 to compromise between speed and accuracy
-    expfac = 4.0
-    dint = jnp.arange(image.array.shape[0] * expfac) / expfac + 1.0
-    fluxes = jnp.interp(dint, d, fluxes)
-    msk = fluxes <= thresh
+    d = jnp.arange(min(image.array.shape[0], image.array.shape[1])) + 1.0
```
ismael-mendoza, line 1185: Could you explain the `+ 1.0` here?

jax_galsim/interpolatedimage.py
```diff
@@ -1247,6 +1235,10 @@ def _find_maxk(kim, max_maxk, thresh):
     # maxk from the image (computed by _inner_comp_find_maxk)
     # by max_maxk from above
     return jnp.minimum(
-        _inner_comp_find_maxk(kim.array, thresh, kx, ky),
+        # jax-galsim tends to be less conservative for maxk
+        # since compared to galsim, it does NOT require 5 rows
+        # of pixels in a row below the threshold.
+        # thus we add some pixels here to ensure the galsim tests pass.
```
ismael-mendoza: you only add one in the end no? 
```suggestion
        # thus we add one pixel here to ensure the galsim tests pass.
```

tests/jax/test_interpolatedimage_utils.py
```diff
@@ -233,8 +233,9 @@ def test_interpolatedimage_utils_comp_to_galsim(
         x_interpolant=x_interp,
     )
 
-    np.testing.assert_allclose(gii.stepk, jgii.stepk, rtol=0.5, atol=0)
-    np.testing.assert_allclose(gii.maxk, jgii.maxk, rtol=0.5, atol=0)
+    np.testing.assert_allclose(jgii.stepk, gii.stepk, rtol=0, atol=1e-6)
+    np.testing.assert_allclose(jgii.maxk, gii.maxk, rtol=0.5, atol=0)
```
ismael-mendoza, line 238:
```suggestion
    #FIXME: match maxk
    np.testing.assert_allclose(jgii.maxk, gii.maxk, rtol=0.5, atol=0)
```

ismael-mendoza (commented):
Just a few questions and some changes. If you think the first two are actually a bug, not sure if these should go into a different PR...?

jax_galsim/interpolatedimage.py
beckermr: yes but I just guess and checked on what worked OK. there is no special reason one is the right answer.

jax_galsim/interpolatedimage.py
beckermr: Yes good catch. I have added that.

beckermr: We only need the minor for the force_maxk branch with the switch to the method above. Good question.

jax_galsim/interpolatedimage.py
beckermr: done!

jax_galsim/interpolatedimage.py
beckermr: added a comment.

jax_galsim/interpolatedimage.py
beckermr: I am pretty sure this code branch was actually never executed due to the parent class having the same branch and thus skipping this call. I am double checking now.

beckermr:
Let's see if removing this unused code works OK. I think it might and I do not recall why or what I was trying to do before.

beckermr:
OK @ismael-mendoza! Thank you for the very helpful code review! This is ready for another look.

jax_galsim/interpolatedimage.py
```diff
@@ -222,14 +219,19 @@ def _maxk(self):
         if self._jax_aux_data["_force_maxk"] > 0:
             return self._jax_aux_data["_force_maxk"]
         else:
-            return super()._maxk
+            # galsim uses a different way to handfle the WCS effects on maxk
```
ismael-mendoza:
```suggestion
            # galsim uses a different way to handle the WCS effects on maxk
```

beckermr:
@ismael-mendoza I recalled the general reason for the two classes here and added some notes on it.

ismael-mendoza (approved):
Thanks LGTM!

PR #224, opened 2026-04-27
feat: investigate maxk differences and add galsim algorithm for maxk

### LINKED ISSUE #189: some numeric thresholds are calculated differently from galsim

### LINKED ISSUE #225: do not set maxk in tests
It turns out this is not needed either for the simpler profiles we use.

beckermr:
I have developed a scan-based implementation of galsim's algorithm for maxk. This algorithm is ~10-20% slower than the one we use now. In most cases it does not make a difference which one we use, so I am opting for the faster one.

I also went through some of the changes in the test suite for maxk and adjusted the threshold for one test for Moffat profiles. I also enabled more of the Moffat tests to be run.

closes #225
closes #189 

For associated changes to the test suite see: https://github.com/GalSim-developers/GalSim-for-JAX-GalSim-Testing/pull/4

### DIFF
--- a/jax_galsim/interpolatedimage.py
+++ b/jax_galsim/interpolatedimage.py
@@ -1214,7 +1214,86 @@ def _inner_comp_find_maxk(arr, thresh, kx, ky):
             -jnp.inf,
         )
     )
-    return jnp.maximum(max_kx, max_ky)
+    # galsim adds one pixel at the end so that maxk is
+    # the k value where things do not pass the threshold,
+    # so we do that here too.
+    return jnp.maximum(max_kx, max_ky) + kx[0, 1] - kx[0, 0]
+
+
+# I am keeping this bit of dead code for posterity.
+# It uses a memory-intensive algorithm that matches galsim's
+# approach for maxk. In practice, the direct search
+# above has MUCH better performance. The code below won't even run
+# on my laptop for parts of the galsim test suite due to memory
+# use. - MRB
+# @jax.jit
+# def _inner_comp_find_maxk(a, thresh, kx, ky):
+#     # Galsim searches the boundary pixels of increasing squares
+#     # at a given distance. It finds the location of the start of
+#     # the first five boundaries where none of the pixels are above
+#     # threshold. To do this operation in JAX, we first compute
+#     # a cumulative version which is the number of pixels within a
+#     # distance that are above threshold. Then we search for the
+#     # location of the first string of five adjacent distances where the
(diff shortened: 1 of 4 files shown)


jax_galsim/interpolatedimage.py
```diff
@@ -1197,24 +1197,91 @@ def _calculate_size_containing_flux(image, thresh):
...
+    # step 1. compute the diff of adjacent elements
+    # The function jnp.diff returns an array of size one less than
+    # the input. So we concatenate a zero at the front. This makes
+    # sense since if the original array is all constant, then the
+    # location of the first five zeros is at the start of the array.
+    delta_msk = jnp.concatenate(
+        [jnp.array([0], dtype=int), jnp.diff(msk)],
+        axis=0,
+        dtype=int,
     )
```
ismael-mendoza: This seemed to fix it for me, although not sure I fully understand why. Specifically, it doesn't only introduce a "one-off" error, as I tried just changing `sind` at the end first and different tests kept breaking...

```suggestion
    delta_msk = jnp.diff(msk).astype(int)
```

beckermr: ~There is actually a bug in galsim and the code above is right. More details later.~ 

Edit: galsim does have a bug, but it appears to not be the cause of issues here. Instead there is a bug here too.

beckermr:
OK This Pr is ready to go. It documents a bunch of work I've done to investigate how we set maxk for interpolated images. TL;DR; what we do now matches galsim exactly in most situations. We can write an algorithm that does exactly what galsim does, but it is more expensive and so doesn't seem worth the trouble. I have left the new algorithm commented out in the code. We can switch to it in the future if we feel it is important.

beckermr:
@ismael-mendoza this one is ready to go!

ismael-mendoza (approved):
LGTM!

PR #229, opened 2026-05-04
fix: use slower maxk comp to match galsim

beckermr:
See PR https://github.com/GalSim-developers/GalSim-for-JAX-GalSim-Testing/pull/5


beckermr:
@ismael-mendoza with the galsim maxk bug fixed, we now need to more complicated maxk comp to pass the test suite. This PR updates the test suite and switches to that.

PR #231, opened 2026-05-07
fix: do not draw from pixel for InterpolatedImage photon shooting

beckermr:
This PR fixes a bug for photon shooting for interpolated images. I had coded it up to draw uniformly from the pixel of the interpolated image and then convolve in the interpolant kernel. That is wrong and instead it should simply treat the original image as a set of delta functions.

Working on comment https://github.com/GalSim-developers/GalSim-for-JAX-GalSim-Testing/pull/1/changes#r3028386824


beckermr:
cc @ismael-mendoza

PR #235, opened 2026-05-09
doc: include conda install command in installation guide

beckermr:
Added conda installation instructions for jax-galsim.


ismael-mendoza:
Thanks @beckermr! Does the conda package get automatically updated everytime we make a new release in pypi?

beckermr:
yes it will, but at a slower timescale

PR #238, opened 2026-05-12
fix: round properly when rendering int images

beckermr:
JAX has platform dependent rounding, so we round everything explicitly.


beckermr:
xref: https://github.com/GalSim-developers/GalSim-for-JAX-GalSim-Testing/pull/11

Issue #237, opened 2026-05-12
better runtime checking of inits


beckermr:
We should probably use equinox to do better runtime checking of values. For example, if you seed a Poisson deviate with a negative mean, the RNG returns -1 with every call.

```
In [6]: pd()
Out[6]: Array(-1., dtype=float32)

In [7]: pd()
Out[7]: Array(-1., dtype=float32)

In [8]: pd()
Out[8]: Array(-1., dtype=float32)

In [9]: pd()
Out[9]: Array(-1., dtype=float32)
```

ismael-mendoza:
that seems like a good idea and it doesn't seem too hard to make it work using this: https://docs.kidger.site/equinox/api/errors/#equinox.error_if

PR #242, opened 2026-05-14
fix: raise errors for invalid shears and PixelScale WCS inits

beckermr:
This PR adds runtime init error checking to the Shear class. It uses the `equinox.error_if` function. The net result is that the code now raises for invalid shears. Unfortunately there is no way to raise exceptions besides the `equinox.EquinoxRuntimeError` and according to the equinox issue tracker, enabling custom exceptions is not a feature they plan to support. We can't use subclassing IIUIC because the inheritance goes the wrong way.

xref: #237

### DIFF
--- a/jax_galsim/shear.py
+++ b/jax_galsim/shear.py
@@ -12,9 +12,9 @@ from jax_galsim.core.utils import ensure_hashable, implements
 @implements(
     _galsim.Shear,
     lax_description=(
-        "While the JAX-GalSim implementation of ``Shear`` will do range checking of "
-        "the shear upon construction, it raises ``equinox.EquinoxRuntimeError`` exceptions "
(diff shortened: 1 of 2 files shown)


beckermr:
OK @ismael-mendoza! This is the first PR in a series that adds runtime init checks to the various classes. I don't think I'll request review for all of them, but I wanted to get eyes on the first one so we all understand the syntax and implications. Please take a look!

beckermr:
closing in favor of #243

ismael-mendoza:
Thanks @beckermr ! Apologies I just noticed this PR, would you like me to review #243 instead?

beckermr:
I'll bump you on a bigger PR. I learned more and needed to change more things.

PR #244, opened 2026-05-15
[WIP] fix: bounds isDefined doesn't vmap properly

beckermr:
This attribute doesn't vmap and is used in a bunch of python control flow. Let's fix that.


beckermr:
closing in favor of #250

PR #247, opened 2026-05-17
fix: version for stable

beckermr:
Replace version import method with importlib.metadata.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -31,7 +31,7 @@ copyright = "2026, GalSim Developers"
 
 from importlib.metadata import version as get_version
 release: str = get_version("jax-galsim")
-
+version = release
 # try:
 #     from jax_galsim._version import version
 # except ImportError:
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -28,10 +28,14 @@
...
+
+from importlib.metadata import version as get_version
+release: str = get_version("jax-galsim")
+
+# try:
+#     from jax_galsim._version import version
+# except ImportError:
+#     version = "0.0.1.dev0"
 
 release = version
```
beckermr:
```suggestion
# 
# release = version
```

docs/conf.py
```diff
@@ -28,12 +28,16 @@
...
 
-try:
-    from jax_galsim._version import version
-except ImportError:
-    version = "0.0.1.dev0"
 
-release = version
+from importlib.metadata import version as get_version
+release: str = get_version("jax-galsim")
+
```
beckermr:
```suggestion
version = release
```

PR #243, opened 2026-05-14
fix: raise errors on invalid inits, failed integrations, etc.

beckermr:
As it says above...

PR #248, opened 2026-05-18
fix: clean up type handling

### LINKED ISSUE #246: clean up and unify type handling
The code now has a large selection of functions that do various type checks and casts. This handling is fragile and confusing. With the adoption of equinox for error handling, we can maybe clean this up  and remove the fragile testing for tracers.

Specifically we should only handle four cases:

- tests of python floats, ints, or complex values in python 
- tests for none
- tests for specific classes from galsim/jax-galskm
- tests of array-like or tracer-like values using equinox runtime checks 

We should also remove the various casts being done in places and instead force more strict typing.

beckermr:
This PR cleans up the type handling.

The goal is to have objects do one of three things:

1. If it is a python numeric type (i.e., `int` or `float`) or a numpy scalar type (i.e., `isinstance(x, np.number)`, `isinstance(x, np.integer)`, etc.; see https://numpy.org/doc/stable/reference/arrays.scalars.html#built-in-scalar-types), convert it to a python type of the appropriate kind.
2. Otherwise, cast to the correct type via `jax.numpy.astype(x, ...)`.
3. For outputs in FITS headers, we allow conversions of numpy/jax arrays to python numeric types as long as there is one element in the array (i.e., it is a numpy scalar type, an array scalar, or a 1D array with one element).

This set of rules is simple to understand, consistently handles array scalars (i.e., arrays with zero dimensions) in that any numpy array scalar is converted to a jax array scalar, and transparently handles JAX tracing.

The diff on the PR results in fewer lines of code, which is a nice side effect!

TODO:
- [x] instead of duck-typing via `try...except` blocks, explicitly test for types
- [x] eliminate the `has_tracers` function in favor of transparent handling
- [x] eliminate use of `cast_to_python_float` / `cast_to_python_int` outside of FITS handling
- [x] document the ideas above in docs

closes #246


jax_galsim/photon_array.py
```diff
@@ -91,19 +91,16 @@ def __init__(
...
-                raise GalSimValueError(
-                    f"The given photon array size {N} is larger than "
-                    f"the allowed total size {_JAX_GALSIM_PHOTON_ARRAY_SIZE}."
-                )
+        self._Ntot = cast_to_int(self._Ntot)
+
+        if _JAX_GALSIM_PHOTON_ARRAY_SIZE is not None and (
+            self._Ntot > _JAX_GALSIM_PHOTON_ARRAY_SIZE
+        ):
+            raise GalSimValueError(
```
beckermr:
```suggestion
            raise ValueError(
```

jax_galsim/photon_array.py
```diff
@@ -91,19 +91,16 @@ def __init__(
...
+        self._Ntot = cast_to_int(self._Ntot)
+
+        if _JAX_GALSIM_PHOTON_ARRAY_SIZE is not None and (
+            self._Ntot > _JAX_GALSIM_PHOTON_ARRAY_SIZE
+        ):
+            raise ValueError(
+                f"The given photon array size {self._Ntot} is larger than "
+                f"the allowed total size {_JAX_GALSIM_PHOTON_ARRAY_SIZE}."
+            )
+
```
beckermr:
```suggestion
```

jax_galsim/angle.py
```diff
@@ -210,11 +209,11 @@ def __mul__(self, other):
     def __div__(self, other):
         if isinstance(other, AngleUnit):
             return self._rad / other.value
-        elif has_tracers(other) or isinstance(other, NON_COMPLEX_TYPES):
+        elif not isinstance(other, Angle):
             return _Angle(self._rad / other)
         else:
             raise TypeError(
-                "Cannot divide Angle by %s of type %s" % (other, type(other))
+                "Cannot multiply Angle by %s of type %s" % (other, type(other))
```
beckermr:
```suggestion
                "Cannot divide Angle by %s of type %s" % (other, type(other))
```

beckermr:
@ismael-mendoza This PR is based on top of #243. I plan to merge it after #243. Comments welcome!

docs/sharp-bits.rst
```diff
@@ -61,6 +61,34 @@ does not affect the original.
    # JAX-GalSim — real_part is a copy
    real_part = complex_image.real  # independent array
 
+Scalar Types, Array Types, and Casting
+--------------------------------------
+
+With the use of JAX, there are now many possible types for numeric data. These include
+
+- **Python scalars**: Things with types that are ``float``, ``int``, or ``complex``.
+- **NumPy scalars**: Things with types that are subclasses are ``np.floating`` and ``np.integer``.
```
ismael-mendoza:
```suggestion
- **NumPy scalars**: Things with types that are subclasses of ``np.floating`` and ``np.integer``.
```

docs/sharp-bits.rst
```diff
@@ -61,6 +61,34 @@ does not affect the original.
    # JAX-GalSim — real_part is a copy
    real_part = complex_image.real  # independent array
 
+Scalar Types, Array Types, and Casting
+--------------------------------------
+
+With the use of JAX, there are now many possible types for numeric data. These include
+
+- **Python scalars**: Things with types that are ``float``, ``int``, or ``complex``.
```
ismael-mendoza: Consider 'Things' -> 'Objects' in this bullet point list?

docs/sharp-bits.rst
```diff
@@ -61,6 +61,34 @@ does not affect the original.
...
+- **NumPy scalars**: Things with types that are subclasses are ``np.floating`` and ``np.integer``.
+- **Numpy array scalars**: Things with a type that is ``np.ndarray`` and has ``np.ndim(...) == 0``.
+- **Numpy arrays**: Things with a type that is ``np.ndarray`` and has ``np.ndim(...) > 0``.
```
beckermr:
```suggestion
- **NumPy array scalars**: Things with a type that is ``np.ndarray`` and has ``np.ndim(...) == 0``.
- **NumPy arrays**: Things with a type that is ``np.ndarray`` and has ``np.ndim(...) > 0``.
```

jax_galsim/image.py
```diff
@@ -333,7 +315,9 @@ def _get_xmin_ymin(array, kwargs, check_bounds=True):
 
     def __repr__(self):
         s = "galsim.Image(bounds=%r" % self.bounds
-        if self.bounds.isDefined() and not has_tracers(self.array):
+        if self.bounds.isDefined() and isinstance(
+            self.array, (np.ndarray, jnp.ndarray, jax.Array)
```
beckermr, line 319:
```suggestion
            self.array, (np.ndarray, jax.Array)
```

docs/sharp-bits.rst
```diff
@@ -61,6 +61,34 @@ does not affect the original.
...
+--------------------------------------
+
+With the use of JAX, there are now many possible types for numeric data. These include
+
+- **Python scalars**: Things with types that are ``float``, ``int``, or ``complex``.
+- **NumPy scalars**: Things with types that are subclasses are ``np.floating`` and ``np.integer``.
+- **NumPy array scalars**: Things with a type that is ``np.ndarray`` and has ``np.ndim(...) == 0``.
+- **NumPy arrays**: Things with a type that is ``np.ndarray`` and has ``np.ndim(...) > 0``.
+- **JAX array scalars**: Things with a type that is ``jnp.ndarray`` and has ``np.ndim(...) == 0``.
+- **JAX arrays**: Things with a type that is ``np.ndarray`` and has ``np.ndim(...) > 0``.
```
beckermr:
```suggestion
- **JAX array scalars**: Things with a type that is ``jax.numpy.ndarray`` and has ``jax.numpy.ndim(...) == 0``.
- **JAX arrays**: Things with a type that is ``jax.numpy.ndarray`` and has ``jax.numpy.ndim(...) > 0``.
```

jax_galsim/image.py
```diff
@@ -333,7 +315,9 @@ def _get_xmin_ymin(array, kwargs, check_bounds=True):
 
     def __repr__(self):
         s = "galsim.Image(bounds=%r" % self.bounds
-        if self.bounds.isDefined() and not has_tracers(self.array):
+        if self.bounds.isDefined() and isinstance(
+            self.array, (np.ndarray, jax.Array)
```
beckermr:
```suggestion
            self.array, (np.ndarray, jnp.ndarray, jax.Array)
```

docs/sharp-bits.rst
```diff
@@ -61,6 +61,34 @@ does not affect the original.
...
+With the use of JAX, there are now many possible types for numeric data. These include
+
+- **Python scalars**: Things with types that are ``float``, ``int``, or ``complex``.
+- **NumPy scalars**: Things with types that are subclasses are ``np.floating`` and ``np.integer``.
+- **NumPy array scalars**: Things with a type that is ``np.ndarray`` and has ``np.ndim(...) == 0``.
+- **NumPy arrays**: Things with a type that is ``np.ndarray`` and has ``np.ndim(...) > 0``.
+- **JAX array scalars**: Things with a type that is ``jax.numpy.ndarray`` and has ``jax.numpy.ndim(...) == 0``.
+- **JAX arrays**: Things with a type that is ``jax.numpy.ndarray`` and has ``jax.numpy.ndim(...) > 0``.
+
+**JAX does not have pure scalar types like NumPY. JAX uses array scalars for those instead.**
```
ismael-mendoza:
```suggestion
**JAX does not have pure scalar types like NumPy. JAX uses array scalars for those instead.**
```

docs/sharp-bits.rst
```diff
@@ -249,11 +260,11 @@ These differences are typically at the level of floating-point round-off
 should not affect scientific conclusions.
 
 ⚠️ Additional Sharp Bits
---------------------------
+------------------------
 
 In the :doc:`api/index` you will find **🔪 JAX-GalSim - The Sharp Bits 🔪** blocks highlighting additional important caveats for specific classes and or methods. These could include things like:
 
-- Many classes do not perform some of Galsim's test for correctness during initialization (e.g., :meth:`~jax_galsim.GSObject.drawImage`).
+- Some classes do not perform some of Galsim's test for correctness during initialization (e.g., :meth:`~jax_galsim.GSObject.drawImage`).
```
ismael-mendoza: I just noticed this comment refers to a class but the example is a method, perhaps we can change it to InterpolatedImage instead where I believe there are still some checks during initialization that we don't do.

```suggestion
- Some classes do not perform some of Galsim's test for correctness during initialization (e.g., :meth:`~jax_galsim.InterpolatedImage`).
```

jax_galsim/angle.py
```diff
@@ -199,7 +198,7 @@ def __sub__(self, other):
         return _Angle(self._rad - other._rad)
 
     def __mul__(self, other):
-        if not (has_tracers(other) or isinstance(other, NON_COMPLEX_TYPES)):
+        if isinstance(other, (Angle, AngleUnit)):
```
ismael-mendoza, line 201: Could you explain why this is the check here? I guess more specifically, why not ensure that `other` is scalar instead?

jax_galsim/bounds.py
```diff
@@ -505,24 +497,25 @@ def __init__(self, *args, **kwargs):
...
-        self.deltax = cast_to_python_float(self.deltax)
-        self.deltay = cast_to_python_float(self.deltay)
+        self.deltax = cast_to_float(self.deltax)
+        self.deltay = cast_to_float(self.deltay)
         if (self.deltax != int(self.deltax)) or (self.deltay != int(self.deltay)):
             raise TypeError("BoundsI must be initialized with integer values")
-        self.deltax = int(cast_to_int(self.deltax))
-        self.deltay = int(cast_to_int(self.deltay))
+        self.deltax = cast_to_int(self.deltax)
+        self.deltay = cast_to_int(self.deltay)
```
ismael-mendoza, line 507: so if I understand correctly, if someone tries to have non static deltax or deltay they will now get a `ConcretizationTypeError,` with no error raised from JAX Galsim explicitly. Is that the intention since we warn the user in the LAX docs and we want to avoid using `has_tracers`?

jax_galsim/image.py
```diff
@@ -744,37 +721,60 @@ def __setitem__(self, *args):
...
-                raise _galsim.GalSimIncompatibleValuesError(
-                    "hermitian == 'x' requires bounds.xmin == 0",
-                    hermitian=hermitian,
-                    bounds=bounds,
-                )
+            self._bounds = _raise_if_nonzero(
+                self.bounds, "x", "hermitian == 'x' requires self.bounds.xmin == 0"
+            )
+            bounds = _raise_if_nonzero(
+                bounds, "x", "hermitian == 'x' requires self.bounds.xmin == 0"
```
ismael-mendoza:
```suggestion
                bounds, "x", "hermitian == 'x' requires bounds.xmin == 0"
```

jax_galsim/image.py
```diff
@@ -744,37 +721,60 @@ def __setitem__(self, *args):
...
-                raise _galsim.GalSimIncompatibleValuesError(
-                    "hermitian == 'y' requires bounds.ymin == 0",
-                    hermitian=hermitian,
-                    bounds=bounds,
-                )
+            self._bounds = _raise_if_nonzero(
+                self.bounds, "y", "hermitian == 'y' requires self.bounds.ymin == 0"
+            )
+            bounds = _raise_if_nonzero(
+                bounds, "y", "hermitian == 'y' requires self.bounds.ymin == 0"
```
ismael-mendoza:
```suggestion
                bounds, "y", "hermitian == 'y' requires bounds.ymin == 0"
```

jax_galsim/photon_array.py
```diff
@@ -91,19 +91,8 @@ def __init__(
         _nokeep=None,
     ):
         self._Ntot = _JAX_GALSIM_PHOTON_ARRAY_SIZE or N
-        if _JAX_GALSIM_PHOTON_ARRAY_SIZE is not None:
```
ismael-mendoza, line 94: I'm not sure I follow why you removed this block here?

jax_galsim/wcs.py
```diff
@@ -1,18 +1,45 @@
...
+# and should never be done anywhere else in the code base
+def _cast_to_static_numeric_scalar(x, msg=None):
+    if isinstance(x, (int, float, np.integer, np.floating)):
+        return x
+
+    if isinstance(x, (np.ndarray, jax.Array, jnp.ndarray)):
+        if x.ndim == 0:
+            return x.item()
+
+        if all(sv for sv in x.shape == 1):
```
ismael-mendoza: I don't think that `x.shape == 1` is an iterable in general for a given numpy/jax.numpy array?

ismael-mendoza (changes requested):
Thanks Matt! I just have a few minor comments/questions and perhaps one small bug in `_cast_to_static_numeric_scalar`?

jax_galsim/bounds.py
beckermr: Yes that is the intention. I am further cleaning this up in https://github.com/GalSim-developers/JAX-GalSim/pull/250.

jax_galsim/angle.py
beckermr: So the goal in this PR is to let arrays and tracers pass through without trying to detect which is which. We can do that for the most part (except for a few cases like BoundsI). So the goal of the check here is guard against common errors of handling Angles and AngleUnits.

jax_galsim/wcs.py
beckermr: Yes this was a bug. Fixed!

docs/sharp-bits.rst
beckermr: done!

jax_galsim/photon_array.py
beckermr: IDK why I removed this one. I just added it back.

beckermr:
pre-commit.ci autofix

beckermr:
OK @ismael-mendoza this one is ready for another look!

ismael-mendoza (approved):
Looks great!

PR #249, opened 2026-05-20
fix: ensure bounds vmaps properly and behaves correctly within JIT

beckermr:
As it says on the tin...


beckermr:
closing in favor of #250

Issue #251, opened 2026-05-21
JAX Galsim Chromatic PSF Implementation


MaxRonce:
I’m currently working with @EiffL on re-implementing the chromatic PSF functionality from the original GalSim in JAX-GalSim.

For now, we are focusing on the case where the SED and morphology are modeled separately. Once this part is in place, we may also extend the implementation to support the other chromatic PSF cases.

beckermr:
Cool. Posting a draft PR for comments as soon as possible would be very useful.

MaxRonce:
@beckermr opened the draft pr https://github.com/GalSim-developers/JAX-GalSim/pull/253

PR #253, opened 2026-05-22
FEAT - Add chromatic psf and the corresponding tests (Issue #251)

MaxRonce:
Should go with the issue https://github.com/GalSim-developers/JAX-GalSim/issues/251

python -m pytest -c /home/maxime/src/galsim/JAX-GalSim_chromaticPSF/pyproject.toml tests/GalSim/tests/test_sed.py tests/GalSim/tests/test_bandpass.py tests/GalSim/tests/test_chromatic.py -q

Tests seams to pass and results are similar to original galsim

image

Element by element testing and benchmark was done using this script :

[bench_dsps_chromatic.py](https://github.com/user-attachments/files/28109502/bench_dsps_chromatic.py)

image

The 10e-5 error on the bottom left plot seams to be only numerical bias due to the NWAVES for the fft and not an actuel implementation bias

Some class diagram before and after :

## Regular Galsim

image

## Jax Galsim

image

workflow


MaxRonce:
# Know issue

# `ChromaticSum` incorrectly treated as separable when summing separable components with different spatial profiles


`ChromaticSum` currently marks itself as separable when all child components are separable:

```python
self._separable = all(o._separable for o in self.obj_list)
```

This is not generally correct. A sum of separable chromatic objects is not necessarily separable.

Ex : physical bulge + disk model:

```text
disk(x, y) * SED_disk(lambda) + bulge(x, y) * SED_bulge(lambda)
```

is not separable in general, even though each component is individually separable. It is separable only in special cases, e.g. if the spatial profiles are identical or if the SEDs are identical up to a scalar factor.

## Observed behavior

When building a model like:

```python
disk = disk_profile * sed_disk
bulge = bulge_profile * sed_bulge
source = disk + bulge
model = ChromaticConvolution([source, psf])
image = model.drawImage(bandpass, ...)
```

`source` becomes a `ChromaticSum`.

Because both `disk` and `bulge` are separable, `ChromaticSum` sets:

```python
self._separable = True
```

Then `ChromaticConvolution.drawImage()` takes the optimized separable path and calls:

```python
o._sed_value(wave)
```

on the `ChromaticSum`.

But `ChromaticSum` does not implement `_sed_value()`, so the base implementation raises:

```python
NotImplementedError
```

## Expected behavior

A sum of separable chromatic components should only be marked separable if the sum itself can be written as:

```text
spatial_profile(x, y) * spectral_weight(lambda)
```

A generic bulge + disk model cannot be represented this way.

### Fixes ??

Always treat `ChromaticSum` as non-separable:

```python
class ChromaticSum(ChromaticObject):
    _separable = False
```

or inside `__init__`:

```python
self._separable = False
```

This is simple and robust, but may lose optimization opportunities for rare truly separable sums.

### More precise fix

Only mark `ChromaticSum` as separable when the sum can actually be factored into:

```text
g(x, y) * h(lambda)
```

That would require implementing compatible `_sed_value()` and `_static_spatial_profile()` behavior only for cases where the children share an equivalent spatial profile or equivalent SED structure.

However, for a generic bulge + disk model with distinct spatial profiles and distinct SEDs, `ChromaticSum` should remain non-separable.

## Current fix

For testing purpose I used this solution :

```text
(disk + bulge) (*) psf = disk (*) psf + bulge (*) psf
```

In code:

```python
disk_image = ChromaticConvolution([disk, psf]).drawImage(bandpass, ...)
bulge_image = ChromaticConvolution([bulge, psf]).drawImage(bandpass, ...)
image = disk_image + bulge_image
```

This is mathematically correct, but it would be preferable for `ChromaticSum` / `ChromaticConvolution` to handle the generic case properly

docs/api/chromatic.rst
```diff
@@ -0,0 +1,83 @@
+Chromatic Profiles
+==================
```
beckermr:
```suggestion
Wavelength-dependent Profiles
=============================
```

Let's use this title to match the upstream documentation. The galsim docs further use separate pages for the SEDS, Bandpasses, and Chromatic objects, but we can skip that bit.

jax_galsim/bandpass.py
```diff
@@ -0,0 +1,192 @@
+"""Bandpass filter representation for chromatic simulations."""
+
+import jax.numpy as jnp
+from jax.tree_util import register_pytree_node_class
+
+
+@register_pytree_node_class
+class Bandpass:
```
beckermr, line 20: Let's not repeat doc strings from upstream and instead use the `implements` decorator, with any caveats put into the `lax_description` keyword.

jax_galsim/bandpass.py
```diff
@@ -0,0 +1,192 @@
...
+        >>> bp = Bandpass(wave, thru)
+    """
+
+    def __init__(self, wave, throughput, blue_limit=None, red_limit=None):
+        self._wave = jnp.asarray(wave, dtype=float)
+        self._throughput = jnp.asarray(throughput)
+
+        if self._wave.ndim != 1 or len(self._wave) < 2:
+            raise ValueError("wave must be a 1-D array with at least 2 elements.")
+        if len(self._throughput) != len(self._wave):
```
beckermr:
```suggestion
        if self._throughput.shape != self._wave.shape:
```

It may be more robust to check the shape directly?

jax_galsim/bandpass.py
```diff
@@ -0,0 +1,192 @@
...
+            raise ValueError("throughput must have the same length as wave.")
+
+        self._blue_limit = float(blue_limit) if blue_limit is not None else float(self._wave[0])
+        self._red_limit = float(red_limit) if red_limit is not None else float(self._wave[-1])
```
beckermr: There is a utility function `cast_to_float` in `jax_galsim.core.utils`. I'd use that here.

jax_galsim/bandpass.py
```diff
@@ -0,0 +1,192 @@
...
+        # Precompute effective wavelength at construction time so it is a
+        # concrete Python float and can be used as a static value under JIT.
+        _w = jnp.linspace(self._blue_limit, self._red_limit, 512)
+        _t = jnp.interp(_w, self._wave, self._throughput)
+        _norm = jnp.trapezoid(_t, _w)
+        self._effective_wavelength_val = (
+            float(jnp.trapezoid(_w * _t, _w) / _norm)
+            if float(_norm) > 0
+            else float(0.5 * (self._blue_limit + self._red_limit))
+        )
```
beckermr: Here `self._throughput` could be a traced array in which case this precompute step won't work.

jax_galsim/bandpass.py
```diff
@@ -0,0 +1,192 @@
...
+        if len(self._throughput) != len(self._wave):
+            raise ValueError("throughput must have the same length as wave.")
+
+        self._blue_limit = float(blue_limit) if blue_limit is not None else float(self._wave[0])
+        self._red_limit = float(red_limit) if red_limit is not None else float(self._wave[-1])
+
+        # Precompute effective wavelength at construction time so it is a
+        # concrete Python float and can be used as a static value under JIT.
+        _w = jnp.linspace(self._blue_limit, self._red_limit, 512)
+        _t = jnp.interp(_w, self._wave, self._throughput)
```
beckermr: This interpolation is linear. There is a slightly better akima interpolant that is fast to compute on-the-fly. See `jax_galsim.core.interpolate`.

jax_galsim/sed.py
```diff
@@ -0,0 +1,240 @@
...
+            f"n_wave={len(self._wave)}, redshift={float(self._redshift):.4f})"
+        )
+
+    def __eq__(self, other):
+        if not isinstance(other, SED):
+            return False
+        return (
+            jnp.array_equal(self._wave, other._wave)
+            and jnp.array_equal(self._flux, other._flux)
+            and jnp.array_equal(self._redshift, other._redshift)
```
beckermr:
```suggestion
            & jnp.array_equal(self._flux, other._flux)
            & jnp.array_equal(self._redshift, other._redshift)
```

beckermr (changes requested):
I took a first pass at review. There are a few big structural things that need addressing before we can review it again.

1. All classes, methods, and functions need to be wrapped with the `@implements` decorator. Any changes specific to JAX need to go in the `lax_description` keyword.
2. Please add the new objects to the API tests in `test_api.py`. From my read of the code, many of those API tests will fail.
3. The Bandpass class cannot cache the effective wavelength on init as a python float since the throughput could be traced.
4. The JAX tests for Bandpass don't appear the test any gradients or JIT with respect to throughput and instead always use a tophat. I think this is possibly why the init stuff from 3 works in the tests, but won't work in general.
5. If an array/computation is meant to only be used at compile time, then we should use `numpy` and wrap it in an `ensure_compile_time_eval` block.

beckermr:
pre-commit.ci autofix

tests/jax/test_chromatic_jax.py
beckermr, line 1: This file needs to be renamed to `test_chromatic_jax.py`. Otherwise, pytest barfs.

tests/jax/test_chromatic_jax.py
```diff
@@ -0,0 +1,411 @@
...
+"""
+
+# ruff: noqa: E402,I001
+
+import jax
+import jax.numpy as jnp
+import pytest
+
+# Enable float64 for accuracy
+jax.config.update("jax_enable_x64", True)
```
beckermr: This needs to be removed. We set float64 or float32 globally for different testing regimes.


```suggestion
```

tests/jax/test_chromatic_jax.py
```diff
@@ -0,0 +1,411 @@
...
+
+import jax_galsim as jgal
+from jax_galsim.chromatic import ChromaticAtmosphere, ChromaticConvolution
```
beckermr: We use the `jgs` idiom for importing `jax_galsim`.

```suggestion
import jax_galsim as jgs
from jax_galsim.chromatic import ChromaticAtmosphere, ChromaticConvolution
```

tests/galsim_tests_config.yaml
```diff
@@ -101,13 +103,20 @@ allowed_failures:
...
   - "object has no attribute 'noise'"
-  - "module 'jax_galsim' has no attribute 'SED'"
+  # JAX-native chromatic subset: array-backed SED/Bandpass only.
+  # Full GalSim spectral I/O, expression parsing, LookupTable metadata,
+  # transforms, thinning, photon shooting, and units are not implemented yet.
+  - "SED.__init__() got an unexpected keyword argument 'spec'"
+  - "SED.__init__() got an unexpected keyword argument 'wave_type'"
+  - "Bandpass.__init__() got an unexpected keyword argument 'wave_type'"
+  - "could not convert string to float"
+  - "is not a valid JAX array type"
```
beckermr: These errors should not be ignored. Instead, we need to adjust the JAX code and/or test suite to account for them.

tests/jax/test_chromatic_jax.py
MaxRonce: Done

tests/jax/test_chromatic_jax.py
MaxRonce: Done

tests/jax/test_chromatic_jax.py
MaxRonce: Done

MaxRonce:
I think if made the change you requsted.
We also did a first pass code quality review with @sfarrens

docs/api/chromatic.rst
```diff
@@ -0,0 +1,83 @@
...
+.. code-block:: python
+
+   chromatic_gal = galsim.Gaussian(half_light_radius=0.5) * sed
+   same_object = sed * galsim.Gaussian(half_light_radius=0.5)
+
+The current implementation focuses on differentiable array-backed SEDs,
+array-backed bandpasses, separable chromatic sources, and non-separable
+Gaussian/Moffat atmospheric PSFs.  Full GalSim spectral I/O, lookup-table
+metadata, Airy/OpticalPSF chromatic optics, and photon shooting are still
+outside this JAX-native subset.
```
ismael-mendoza:
```suggestion
metadata, Airy/OpticalPSF chromatic optics, and photon shooting have not been implemented.
```

jax_galsim/bandpass.py
```diff
@@ -0,0 +1,206 @@
...
+
+    @property
+    @implements(_galsim.Bandpass.effective_wavelength)
+    def effective_wavelength(self):
+        return self.calculateEffectiveWavelength()
+
+    @implements(
+        _galsim.Bandpass.calculateEffectiveWavelength,
+        lax_description=(
+            "The JAX implementation returns a traced scalar when throughput is traced."
```
ismael-mendoza: Should also mention in the LAX description that precise flag = True is not supported

tests/galsim_tests_config.yaml
```diff
@@ -101,13 +115,18 @@ allowed_failures:
...
-  - "module 'jax_galsim' has no attribute 'Bandpass'"
   - "module 'jax_galsim' has no attribute 'Refraction'"
   - "module 'jax_galsim' has no attribute 'FRatioAngles'"
   - "module 'jax_galsim' has no attribute 'PupilAnnulusSampler'"
   - "module 'jax_galsim' has no attribute 'TimeSampler'"
   - "object has no attribute 'noise'"
-  - "module 'jax_galsim' has no attribute 'SED'"
+  # JAX-native chromatic subset: array-backed SED/Bandpass only.
+  # Full GalSim spectral I/O, expression parsing, LookupTable metadata,
+  # transforms, thinning, photon shooting, and units are not implemented yet.
```
ismael-mendoza: Comment not needed in this file if already documented in docs/docstrings

tests/galsim_tests_config.yaml
```diff
@@ -31,6 +34,18 @@ enabled_tests:
 # tests to explicitly skip
 # applied on top of the enabled set above
 skipped_tests:
+  galsim:
+    # JAX-GalSim's chromatic subset is array-backed and does not implement
+    # GalSim spectral file/string expression parsing, zeropoints, DCR moment
+    # helpers, chromatic affine transforms, or photon shooting.
```
ismael-mendoza: some comment as below

jax_galsim/sed.py
```diff
@@ -0,0 +1,190 @@
...
+        if self._wave.ndim != 1 or len(self._wave) < 2:
+            raise ValueError("wave must be a 1-D array with at least 2 elements.")
+        if self._flux.shape != self._wave.shape:
+            raise ValueError("flux must have the same shape as wave.")
+
+    @property
+    @implements(
+        None,
+        lax_description="JAX-GalSim-specific static wavelength grid for array-backed SEDs.",
+    )
```
ismael-mendoza: Not sure if these and other should just be a docstring?

jax_galsim/sed.py
```diff
@@ -0,0 +1,190 @@
...
+    lax_description="""\
+JAX-GalSim supports array-backed SEDs only.  The wavelength grid is static
+and the flux array may be traced.  File input, string expressions, units,
+LookupTable metadata, thinning, and magnitude/zeropoint helpers are not
+implemented.
+""",
+)
+@register_pytree_node_class
+class SED:
+    def __init__(self, wave, flux, redshift=0.0):
```
ismael-mendoza: I think we should keep the same API as the original SED class but throw errors if arguments that we do not support are provided

docs/api/chromatic.rst
```diff
@@ -0,0 +1,83 @@
+Wavelength-dependent Profiles
+=============================
+
+.. currentmodule:: jax_galsim
+
+JAX-GalSim supports a JAX-native subset of GalSim chromatic rendering.  The
```
ismael-mendoza:
```suggestion
JAX-GalSim supports a JAX-compatible subset of GalSim chromatic rendering.  The
```
I think this is a bit clearer, can you modify other instances below and docstrings?

tests/conftest.py
```diff
@@ -98,7 +98,12 @@ def pytest_collection_modifyitems(config, items):
         ):
             item.add_marker(skip)
 
-        if any([t in item.nodeid for t in test_config["skipped_tests"]["coord"]]):
+        skipped_tests = test_config["skipped_tests"].get("galsim", []) + test_config[
+            "skipped_tests"
+        ].get("coord", [])
```
ismael-mendoza: Could you modify it so it doesn't get split into multiple lines? It's very hard to read

tests/galsim_tests_config.yaml
```diff
@@ -101,13 +115,18 @@ allowed_failures:
...
   - "module 'jax_galsim' has no attribute 'PupilAnnulusSampler'"
   - "module 'jax_galsim' has no attribute 'TimeSampler'"
   - "object has no attribute 'noise'"
-  - "module 'jax_galsim' has no attribute 'SED'"
+  # JAX-native chromatic subset: array-backed SED/Bandpass only.
+  # Full GalSim spectral I/O, expression parsing, LookupTable metadata,
+  # transforms, thinning, photon shooting, and units are not implemented yet.
+  - "SED.__init__() got an unexpected keyword argument 'spec'"
+  - "SED.__init__() got an unexpected keyword argument 'wave_type'"
+  - "Bandpass.__init__() got an unexpected keyword argument 'wave_type'"
```
ismael-mendoza: This means the API is breaking and skipping most of the galsim tests no? Is there a way to maintain the original galsim API?

tests/conftest.py
```diff
@@ -98,7 +98,12 @@ def pytest_collection_modifyitems(config, items):
...
+        if any(
+            [item.nodeid == t or item.nodeid.startswith(t + "[") for t in skipped_tests]
+        ):
```
ismael-mendoza: could you explain why this modifcation was needed in a comment perhaps with an example?

jax_galsim/bandpass.py
```diff
@@ -0,0 +1,206 @@
...
+    _galsim.Bandpass,
+    lax_description="""\
+JAX-GalSim supports array-backed bandpasses only.  The wavelength grid is
+static and the throughput array may be traced.  File input, string
+expressions, zeropoints, units, and LookupTable metadata are not implemented.
+""",
+)
+@register_pytree_node_class
+class Bandpass:
+    def __init__(self, wave, throughput, blue_limit=None, red_limit=None):
```
ismael-mendoza: The class signature does not match the original GalSim API for BandPass

jax_galsim/bandpass.py
```diff
@@ -0,0 +1,206 @@
...
+            cast_to_float(red_limit)
+            if red_limit is not None
+            else cast_to_float(self._wave[-1])
+        )
+
+    @property
+    @implements(
+        None,
+        lax_description="JAX-GalSim-specific static wavelength grid for array-backed bandpasses.",
+    )
```
ismael-mendoza: if this is not implemented in GalSim should we use the `@implements` decorator or simply use a docstring?

jax_galsim/bandpass.py
```diff
@@ -0,0 +1,206 @@
...
+    def tophat(cls, blue_limit, red_limit, n_wave=100):
+        wave = jnp.linspace(blue_limit, red_limit, n_wave)
+        return cls(wave, jnp.ones(n_wave))
+
+    def tree_flatten(self):
+        children = (self._throughput,)
+        aux_data = {
+            "wave": tuple(self._wave.tolist()),
+            "blue_limit": ensure_hashable(self._blue_limit),
+            "red_limit": ensure_hashable(self._red_limit),
```
ismael-mendoza: Could you explain why you use `ensure_hashable` here? I don't see this pattern in `tree_flatten` anywhere else in the codebase

jax_galsim/bandpass.py
```diff
@@ -0,0 +1,206 @@
...
+    @classmethod
+    def tree_unflatten(cls, aux_data, children):
+        return cls(
+            wave=jnp.asarray(aux_data["wave"], dtype=float),
+            throughput=children[0],
+            blue_limit=aux_data["blue_limit"],
+            red_limit=aux_data["red_limit"],
+        )
+
+    def __repr__(self):
```
ismael-mendoza: I'm surprised `ensure_hashable` is not used here instead

jax_galsim/sed.py
ismael-mendoza: could you let me know if there is a particular reason you could not conform to the original signature?

jax_galsim/chromatic.py
```diff
@@ -0,0 +1,974 @@
...
+    _separable = False
+
+    def __init__(
+        self,
+        fwhm_ref,
+        lam_ref,
+        alpha=-0.2,
+        profile="gaussian",
+        moffat_beta=4.765,
+        gsparams=None,
```
ismael-mendoza: this signature is also different from GalSIM why? Also we should avoid hard coded number defaults

jax_galsim/chromatic.py
```diff
@@ -0,0 +1,974 @@
...
+
+    def __repr__(self):
+        inner = ", ".join(repr(o) for o in self.obj_list)
+        return f"ChromaticConvolution([{inner}])"
+
+
+# ---------------------------------------------------------------------------
+# Monkey-patch GSObject.__mul__ to return a simple chromatic object for SEDs
+# ---------------------------------------------------------------------------
+
```
ismael-mendoza: could you explain why we need this?

jax_galsim/bandpass.py
```diff
@@ -0,0 +1,206 @@
...
+            )
+        in_band = (wave >= self._blue_limit) & (wave <= self._red_limit)
+        return jnp.where(in_band, throughput, 0.0)
+
+    @implements(_galsim.Bandpass.__mul__)
+    def __mul__(self, other):
+        if isinstance(other, Bandpass):
+            wave = jnp.unique(jnp.concatenate([self._wave, other._wave]))
+            t = jnp.interp(wave, self._wave, self._throughput, left=0.0, right=0.0)
+            t2 = jnp.interp(wave, other._wave, other._throughput, left=0.0, right=0.0)
```
ismael-mendoza: could we consider using akima interpolation here too?

PR #258, opened 2026-06-23
fix: better init of spergel HLR and tabulate xValue

beckermr:
This PR has a new way to initialize the spergel HLR and a better way to compute autodiff using implicit differentiation.


jax_galsim/spergel.py
```diff
@@ -165,6 +136,41 @@ def reducedfluxfractionFunc(z, nu, norm):
     return fluxfractionFunc(z, nu, alpha=0.0) / norm
 
 
+# code here is from JAX osurce for testing custom_root
```
beckermr:
```suggestion
# code here is from JAX source for testing custom_root
```

ismael-mendoza:
hello - i started looking into what you requested in slack. 

Regarding the image artifacts with `slen=101` for main and this PR respectively (the plots show the residual between galsim - jax_galsim):

image

image

but using `slen=201` (again main branch and then this PR):

image


image


I used the same code that's [here](https://github.com/GalSim-developers/JAX-GalSim/issues/257#issuecomment-4730989849) to test. I don't understand why the residual changes so much in this PR after doubling the `slen` because GalSim reports that the good size is `52`. Do you have any insight? (although in both cases it's smaller than what we had before so maybe it's ok anyways? )

ismael-mendoza:
re the regression when nu is not constant - I could not find any. Perhaps I misunderstood but here is the code I was using to test this in my laptop. 

```python
import numpy as np
import jax_galsim as jgs

def _draw_spergel_jgs(
    params: dict, nu: float | Array, *, psf: jgs.GSObject, fft_size: int, slen: int
):
    gsparams = jgs.GSParams(minimum_fft_size=fft_size, maximum_fft_size=fft_size)
    prof = jgs.Spergel(nu=nu, flux=params["flux"], half_light_radius=params["hlr"])
    prof = prof.shear(q=params["q"], beta=params["beta"] * jgs.degrees)
    gal = jgs.Convolve([prof, psf]).withGSParams(gsparams)
    return gal.drawImage(nx=slen, ny=slen, scale=0.2, dtype=jnp.float64).array

# let's average the time over 1000 runs each while varying the parameters
# but fixing the PSF


times1 = []
times2 = []

xpsf = jgs.Gaussian(half_light_radius=0.7, flux=1.0)

draw_jax_nu_float = jit(
    partial(_draw_spergel_jgs, nu=-0.6, psf=xpsf, fft_size=256, slen=201)
)
draw_jax_nu_array = jit(partial(_draw_spergel_jgs, psf=xpsf, fft_size=256, slen=201))


for ii in range(1000):
    params = {
        "flux": np.random.uniform(1, 3, size=()).item(),
        "hlr": np.random.uniform(0.4, 0.5, size=()).item(),
        "q": 1.0,
        "beta": 0.0,
    }
    _nu = np.random.uniform(-0.55, -0.65, size=()).item()
    _nu = device_put(jnp.array(_nu))
    params_jax = device_put(params)

    # compilation
    if ii == 0:
        draw_jax_nu_float(params_jax)
        draw_jax_nu_array(params_jax, _nu)

    # timing
    t1 = time.time()
    with transfer_guard("disallow"):
        _ = block_until_ready(draw_jax_nu_float(params_jax))
    t2 = time.time()
    times1.append(t2 - t1)

    t1 = time.time()
    with transfer_guard("disallow"):
        _ = block_until_ready(draw_jax_nu_array(params_jax, _nu))
    t2 = time.time()
    times2.append(t2 - t1)


t_nu_float = np.mean(times1)
t_nu_array = np.mean(times2)
print(f"Average time nu float: {t_nu_float:.4f}")
print(f"Average time nu array: {t_nu_array:.4f}")
```

Did I miss something?

beckermr:
This PR is ready for review @ismael-mendoza!

jax_galsim/spergel.py
```diff
@@ -165,6 +136,41 @@ def reducedfluxfractionFunc(z, nu, norm):
...
+
+    solution, _ = jax.lax.while_loop(cond, body, (low, high))
+    return solution
+
+
+# end of code from jax source
+
+
 @jax.jit
 def calculateFluxRadius(alpha, nu, zmin=0.0, zmax=40.0):
```
ismael-mendoza: is it ok that zmin and zmax are no longer used here? Should they be removed from the function signature?

jax_galsim/spergel.py
```diff
@@ -240,6 +246,23 @@ def _spergel_hlr_pade(x):
     return pm / qm
 
 
+@jax.jit
+def _spergel_hlr_binary_search_plus_pade_init(nu):
+    """Return radius R enclosing flux fraction 0.5 in unit of the scale radius r0"""
+    z = _spergel_hlr_pade(nu)
+    eps = 1e-6
```
ismael-mendoza, line 255: is it ok to hard code this value? what happens if the desired accuracy is not reached?

.github/workflows/benchmarks.yaml
```diff
@@ -40,6 +40,13 @@ jobs:
           python -m pip install --upgrade pip
           python -m pip install .[dev]
 
+      - name: Install glibc debug symbols
```
ismael-mendoza, line 44: could you explain why you added this package?

jax_galsim/spergel.py
```diff
@@ -112,51 +112,22 @@ def fnu(z, nu):
...
+        [msk0, msk1, msk2, msk3, msk4],
         [f0(z), f1(z), f2(z), f3(z), f4(z)],
-        default=fnu(z, nu),
+        default=fnu(nu_safe, z),
     )
 
 
 @jax.jit
-def fz_nu(z, nu):
+def _fz_nu(nu, z):
```
ismael-mendoza, line 122: I'm not sure I follow why you dropped the small z guard here?

ismael-mendoza (changes requested):
Thanks Matt! I am not able to follow/check some of the detailed math calculations, but I did have some high level questions

jax_galsim/spergel.py
```diff
@@ -286,9 +311,17 @@ def __init__(
                     scale_radius=scale_radius,
                 )
             else:
+                # for python floats, we can use galsim on the CPU-side to do this
+                # quickly as long as we ensure it is done at compile time.
+                if isinstance(nu, float):
+                    with jax.ensure_compile_time_eval():
+                        hlr = _galsim.Spergel(nu, scale_radius=1).half_light_radius
+                else:
+                    hlr = _spergel_hlr_binary_search_plus_pade_init(nu)
```
ismael-mendoza, line 327: ah one more comment, do you think it might be good to add some description to the LAX documentation regarding this? or too low level for someone using the code to care about? I'm thinking the different numerics could surprise someone?

.github/workflows/benchmarks.yaml
beckermr: It is needed by the benchmarking tooling. I will add a comment.

jax_galsim/spergel.py
beckermr: This value is not the accuracy. It is the width of the search interval around the initial guess. It needs to be bigger than largest absolute error of the initial guess. I will add a comment.

jax_galsim/spergel.py
beckermr: It makes the function not continuous numerically and the small z values from the other function are fine.

jax_galsim/spergel.py
beckermr: The differences in numerics should not be noticeable, but I will add a note to the LAX docs.

jax_galsim/spergel.py
beckermr: These should be removed yes.

PR #261, opened 2026-07-27
Add DES_PSFEx (PSFEx model reader) to jax_galsim.des

AdamField118:
## Add `DES_PSFEx` to `jax_galsim.des`

Hi! Nice to meet you, this is my first PR ever, happy to adjust anything to match your conventions.

This ports `galsim.des.DES_PSFEx` so a PSFEx model (`*_psfcat.psf`) can be read and evaluated entirely within JAX-GalSim. Right now, if you want an empirical PSF you have to drop back to reference GalSim just to build it; this closes that gap. I hit it while benchmarking a weak lensing pipeline's dataset generation before and after porting to jax-galsim. `DES_PSFEx` was the last thing keeping regular GalSim as a dependency.

### What it does
- `read()` is host-side FITS I/O, unchanged from GalSim (parses the PSFEx header + PCA basis cube).
- `getPSFArray()` is the per-position evaluation of the interpolated PSF. It's reimplemented in JAX so it's `jit`-able, `vmap`-able over image positions, and differentiable w.r.t. the image position. The one non-obvious change: GalSim builds the `[1, x, x^2, ...]` powers with an in-place `np.empty` loop, which tracing forbids, so I use a cumulative product with the same recurrence.
- `getPSF()` wraps the array in an `InterpolatedImage` exactly as GalSim does (`Lanczos(3)`, `scale = PSF_SAMP`) and applies the WCS if one was provided.

### Scope (intentionally narrow)
Just `DES_PSFEx`, not `DES_Shapelet`, and not the config-framework registration (`des_psfex` input type / `DES_PSFEx` object type), since JAX-GalSim doesn't have a config system. That's noted in the `lax_description`. Happy to follow up with more in a separate PR if you'd like it.

### Testing
`tests/jax/test_des_psfex_jax.py` compares against reference GalSim using the existing `DECam_00154912_12_psfcat.psf` in the `tests/GalSim` submodule:
- `getPSFArray` and the drawn effective-PSF image match GalSim to ~single precision (the PSFEx bases are `float32`). Drawn with `method='no_pixel'`, since PSFEx PSFs already include the pixel.
- `getPSFArray` is verified `jit`-able, `vmap`-able across a batch of positions, and differentiable in position (finite gradient).

`ruff check` and `ruff format` are clean.

### Open questions
- Happy to move the module, rename things, or adjust the test tolerances to match your house style.
- Let me know if you'd prefer the config registration included after all, or `DES_Shapelet` alongside it.

### DIFF
--- a/jax_galsim/__init__.py
+++ b/jax_galsim/__init__.py
@@ -102,6 +102,7 @@ from .sensor import Sensor
 from . import bessel
 from . import fits
 from . import integ
+from . import des
 
 # this one is specific to jax_galsim
 from . import core
--- a/jax_galsim/des/__init__.py
+++ b/jax_galsim/des/__init__.py
@@ -1,0 +1,1 @@
+from .des_psfex import DES_PSFEx
--- a/jax_galsim/des/des_psfex.py
+++ b/jax_galsim/des/des_psfex.py
@@ -7,11 +7,14 @@ import os
 import galsim as _galsim
 import galsim.des  # noqa: F401  (populates _galsim.des for @implements below)
 import jax.numpy as jnp
-import numpy as np
 from jax.tree_util import register_pytree_node_class
 
 from jax_galsim._pyfits import pyfits
-from jax_galsim.core.utils import ensure_hashable, implements
+from jax_galsim.core.utils import (
+    cast_numpy_array_to_native_byte_order,
+    cast_to_float,
+    implements,
+)
 from jax_galsim.errors import GalSimIncompatibleValuesError
 from jax_galsim.fits import FitsHeader
 from jax_galsim.image import Image
@@ -24,11 +27,17 @@ The JAX-GalSim version of ``DES_PSFEx`` does not register itself with the
 GalSim config framework (the ``des_psfex`` input type and ``DES_PSFEx`` object
 type are not available), since JAX-GalSim does not implement config processing.
 
-As a PyTree, all of the data read from the PSFEx file (the PCA basis and the
-polynomial fit parameters) is static auxiliary data stored as NumPy arrays,
-and only the ``wcs`` is traced. Autodiff is therefore supported with respect
-to the ``image_pos`` argument (via ``getPSFArray``) and the ``wcs``, but not
-with respect to the fixed PSFEx calibration data.
+As a PyTree, the data read from the PSFEx file (the PCA basis and the
+polynomial zero points and scales) together with the ``wcs`` are traced
+children, so a batch of PSFEx models can be stacked and evaluated in a single
+``jit``/``vmap`` call. Only ``fit_order`` and ``fit_size`` are static
+auxiliary data, since they set the number of polynomial terms and hence the
+shapes of the traced arrays; batching therefore applies to models sharing a
+polynomial degree, which is the usual case within one instrument and reduction.
+
+``file_name`` is deliberately *not* part of the PyTree, so that models read
+from different files share a tree structure and can be batched together. An
+instance rebuilt by ``tree_unflatten`` has ``file_name`` set to ``None``.
 """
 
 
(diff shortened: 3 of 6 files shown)


beckermr (changes requested):
This class needs to be registered as a pytree and we need think about what kinds of autodiff we'd allow.

AdamField118:
I'll register it as a pytree. My idea of the split: basis, the polynomial zero/scale constants, and wcs as traced children (keeping it consistent with how other jax_galsim objects treat numeric params), with fit_order/fit_size as static aux_data since they set the loop unrolling and array shapes. tree_unflatten will rebuild via `object.__new__` to skip the file read, matching what CelestialCoord/Image do.

On autodiff scope: I think the physically meaningful gradient is w.r.t. image_pos (where on the focal plane the PSF is evaluated), which already works and comes through the method argument rather than the pytree. Differentiating w.r.t. basis/wcs/calibration constants falls out structurally once they're leaves, but I don't see a concrete use case for those. I'd be happy to lock any of them into aux_data if you'd rather keep the differentiable surface narrow. What's your preference?

beckermr:
Let's make all of the data read in from the PSF file auxiliary and store it as numpy arrays. The WCS will need to be traced.

AdamField118:
Registered as a pytree with all the PSFEx-file data as static auxiliary numpy arrays and wcs as the only traced child. `tree_unflatten` rebuilds via `object.__new__` so it doesn't re-read the file. I ran the basis through `ensure_hashable` so the metadata stays hashable when an instance is passed as an argument to a jitted/vmapped function. Let me know what you think!

beckermr:
@AdamField118 You'll need to go through the tests and ensure that the upstream galsim ones are running.

AdamField118:
Upstream test_des.py now runs against jax_galsim (1 passed, 4 allowed failure). test_psf previously aborted at galsim.Catalog (upstream catches OSError, but jax_galsim raises AttributeError) before ever reaching the PSFEx code, so I added a scoped conftest fixup for that module; it now exercises DES_PSFEx construction with a WCS, getPSF, getLocalWCS, drawImage(method='no_pixel') and the adaptive-moment comparisons against GalSim's reference values, stopping only at DES_Shapelet. Also added allowlist entries for the unported des members (MultiExposureObject, WriteMEDS, DES_Shapelet).

One thing I thought about that's a bit unrelated: with the basis in aux_data, each distinct PSFEx file is a separate treedef, so batching across different PSF files in one vmap isn't possible. As someone who works with a pipeline that draws multiple galaxies with different PSF files, I would just have to group per-file then vmap in those groups, but I figured I would tell you about this little quirk just because it feels topical.

beckermr:
Excellent point about JITing over PSF models, not just PSF positions. As long as tests for this use case are added, I'd be fine with the PSFEx data in the traced data. Sorry I missed this the first time around!

AdamField118:
Great! I'll move the PSFEx data into the traced children and add a test that JITs/vmaps across multiple PSF models.

You helped me realize that what I meant in my last comment was PSF models not PSF files, but that raises one more design question: `file_name` is currently in `aux_data`, which means two models loaded from different files get different treedefs and can't be stacked into one batch. Do you like the idea of dropping `file_name` from the treedef, making a pytree-reconstructed instance not retain it?

AdamField118:
The PSFEx data and the `wcs` are now traced children. `fit_order`/`fit_size` stay static.

I went ahead and dropped `file_name` from the treedef like I mentioned earlier, so models from different files can stack together. A reconstructed instance just gets `file_name = None`. Let me know if you'd rather put it back.

Added `test_des_psfex_vmap_over_multiple_psf_models`, it stacks two different models and runs them through one `jit(vmap(...))`, checking each matches what that model gives on its own, and that the two give different answers.

beckermr (approved):

PR #264, opened 2026-08-26
Cap the k-space interpolation stencil at one period

### LINKED ISSUE #81: figure out why k-space drawing for the sinc interpolant is so slow
The k-space drawing for the sinc interpolant is really slow. We should figure out why and try to fix it.

AdamField118:
Fixes #81.

Caps the k-space interpolation stencil at one period, using the closed-form
Dirichlet kernel for the sinc interpolant. The k-image is periodic and indices are
already wrapped with `% nkx`, so aliases past one period land on points the stencil
already covers, sinc's `ixrange` is 63662 at the default `kvalue_accuracy`, so a
9x9 draw asked for ~3e11 elements and failed to allocate. It now runs in 0.05s.

Compact interpolants (Nearest, Linear, Cubic, Quintic, Lanczos 3/5) are bitwise
identical before and after. With sinc, `kValue` now agrees with Quintic and
Lanczos(7) on the same image.

One deliberate difference from GalSim: the fast path triggers on `2*xrange <= n`
rather than `xrange() <= n`, since the single-alias shortcut is only valid when the
nearest other alias falls outside `xrange` (the looser test is off by ~1.7e-2 for
Quintic at n=4). `xvalWrapped` has no call sites in GalSim, so it isn't reachable there.

Re-enabled `SincInterpolant()` in `test_interpolatedimage_utils.py`, where it was
commented out.

### DIFF
--- a/jax_galsim/interpolant.py
+++ b/jax_galsim/interpolant.py
@@ -156,6 +156,16 @@ class Interpolant:
     def _xval_noraise(self, x):
         return self.__class__._xval(x)
 
+    def _xval_wrapped_noraise(self, x, n):
+        """The sum of ``xval`` over all aliases ``x + j*n``, as in
+        ``Interpolant::xvalWrapped`` in GalSim."""
+        xdown = x - n * jnp.floor(x / n + 0.5)
+        if 2 * self.xrange <= n:
+            return self._xval_noraise(xdown)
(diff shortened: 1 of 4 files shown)


beckermr (approved):
LGTM! Thank you!

PR #266, opened 2026-09-18
use scatter add in test of drawing full scenes

ismael-mendoza:
@beckermr I add the scatter add approach to the test that draws full scenes which replaces the old scan approach


beckermr (approved):
LGTM and thank you!

